PR #77, opened 2020-07-30
first page no adaptability

varseniuk:
[DEMO LINK](https://varseniuk.github.io/react_phone-catalog/)

- first page no adaptability


public/index.html
```diff
@@ -7,5 +7,388 @@
...
+          </div>
+          <div class="promo__slide">
+            <div class="promo__line promo__line--active"></div>
+            <div class="promo__line"></div>
+            <div class="promo__line"></div>
+          </div>
+
+        </section>
+
+        <section class="hot section">
```
irenhh, line 73: what is hot?

public/index.html
```diff
@@ -7,5 +7,388 @@
...
+            </article>
+          </div>
+        </section>
+
+        <section class="new section">
+          <div class="section__top">
+            <h2 class="section__heading">Brand new models</h2>
+            <div class="scroll-container">
+              <div class="scroll scroll--left"></div>
+              <div class="scroll scroll--right"></div>
```
irenhh, line 246: there should be icons

src/styles/nav.scss
```diff
@@ -0,0 +1,44 @@
...
+
+    &--active {
+      position: relative;
+      color: $primaryColor;
+
+      &:after {
+        position: absolute;
+        content: "";
+        left: 0;
+        bottom: -1px;
```
irenhh, line 37: why -1?

public/index.html
varseniuk: It was meant to be "hot-prices") But there is no actual need in this class, so I removed it.

public/index.html
varseniuk: We've discussed this question today. I'll leave it as is. For now. If you don't mind

src/styles/nav.scss
varseniuk: Because it's according to design

PR #131, opened 2022-12-03
Done

Oleksii-Blyzniuk:
https://oleksii-blyzniuk.github.io/react_phone-catalog/


Anastasiia-Svintsova (changes requested):
- Images position doesn't match the design
image
- Logo should lead to a Home page
image
- This link should lead somewhere
Animation
- Image disappears in the slider on a small screen
Animation
- Also try to pass all tests (At least most of them)
Animation

loralevitska (changes requested):
looks great) Let's improve

1. Better set click on whole card not only on image 
image
2. add transition to image for `hover` effect
3. center icon and word 
image
4. 
image

PR #136, opened 2022-12-30
Actualise data according to design

Bohdan-Maksymiuk:
added public resources from [product_catalog](https://github.com/mate-academy/product_catalog/tree/main/public) to match design
And replaced links in readme


README.md
```diff
@@ -5,8 +5,8 @@
 ## Description
 Implement Products catalog following [this design](https://www.figma.com/file/uEetgWenSRxk9jgiym6Yzp/Phone-catalog-redesign?node-id=1%3A2).
 
-Use [products](https://mate-academy.github.io/react_phone-catalog/api/products.json)
-and [product details](https://mate-academy.github.io/react_phone-catalog/api/products/motorola-xoom.json)
+Use [products](https://mate-academy.github.io/react_phone-catalog/api/phones.json)
+and [product details](https://mate-academy.github.io/react_phone-catalog/api/phones/apple-iphone-7-32gb-black.json)
```
mgrinko, line 9: Why do these links use `/phones` instead of `/products`?
Logically it is wrong, because we can display not only phones

mgrinko (approved):

PR #147, opened 2023-02-14
first try

Vovahunter:
Demo link - https://vovahunter.github.io/react_phone-catalog/

### DIFF
--- a/src/components/Header/Header.scss
+++ b/src/components/Header/Header.scss
@@ -11,7 +11,6 @@
   &__main-container {
     display: flex;
     justify-content: space-between;
-    border: 0 1px solid $color-elements;
     max-width: 1136px;
     margin: 0 auto;
   }


AlisaVlasova (changes requested):
awesome!)
set max-width to header content for large screens, like on footer and main
image

is it desktop only? 
image

Vovahunter:
https://vovahunter.github.io/react_phone-catalog/

src/components/CartCheckout/CartCheckout.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const { cart } = useContext(Context);
+
+  const [isClicked, setIsClicked] = useState(false);
+
+  const cartTotalPrice = () => {
+    const currentItems = JSON.parse(localStorage.getItem('cart') || '[]');
+
+    return currentItems.reduce(
+      // eslint-disable-next-line max-len
+      (sum: number, current: Product) => sum + (current.price - ((current.price / 100) * current.discount)), 0,
```
TarasHoliuk, line 20:
```suggestion
      (sum: number, current: Product) => 
        sum + (current.price - ((current.price / 100) * current.discount)),
      0,
```
don't disable lint. it's better to format it in another way

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const productCategoryList = productCategory.map(item => item.type);
+
+  const searchBarCondition = productCategoryList.some(
+    item => location.pathname.includes(item),
+  ) && !productId;
+
+  const cartCondition = location.pathname.includes('cart');
+  const favoriteCondition = location.pathname.includes('favorite');
+
+  const handleIsOpened = useCallback((menuStatus: boolean) => {
```
TarasHoliuk, line 38: you will always toggle it to the opposit value, so I guess it might be called `toggleMenu`, etc., and toggle it here (shouldn't accept an argument)

src/types/ProductDetails.ts
```diff
@@ -0,0 +1,49 @@
+export interface ProductDetailsType {
```
TarasHoliuk, line 1:
```suggestion
export interface ProductDetails {
```

TarasHoliuk (changes requested):
GJ!
I've added a few minor comments

PR #149, opened 2023-02-16
last saved 17.02

VitalyPasechnik:


sTorba24 (changes requested):
Add demo link please and ask review again :)

VitalyPasechnik:
Hello,

Changed!

Thanks!

sTorba24 (changes requested):
Good work but needed some fixes ;)

1. Here should be 4 cards seems 
image

2.Looks like there are not some styles and icons, had you deployed everything?
image

3.Don't think it should look so :)
image

4.remove additional line above tech + about block should have the same indentation as it title
image

5. Change total value immediately with changing of cart(number of goods, new item). not on click on checkout button
image

PR #153, opened 2023-03-08
Develop

andriimelnyq:
https://andriimelnyq.github.io/react_phone-catalog/


TarasHoliuk:
The layout on this page is broken for some screen widths

image

I'm not 100% sure, but maybe better to display 1 item on the left side?

image

src/styles/utils/_utils.scss
```diff
@@ -0,0 +1,17 @@
+@import "../reset.scss";
+@import "../normalize.scss";
+
+@import "fonts";
+@import "extends";
+@import "variables";
+@import "mixins";
+
+@import "../blocks/main-container.scss";
```
TarasHoliuk, line 9: why `_utils.scss` contains all blocks?

TarasHoliuk: it looks like `main.scss`, etc.

src/components/CartItem/CartItem.tsx
```diff
@@ -0,0 +1,136 @@
...
+import './CartItem.scss';
+
+type Props = {
+  cartItem: CartItemType,
+  cart: CartItemType[],
+  setCart: (cart: CartItemType[]) => void,
+};
+
+export const CartItem: React.FC<Props> = ({ cartItem, cart, setCart }) => {
+  const transformType = (type: string) => {
```
TarasHoliuk, line 14: It may accept the variable with a more specific type
you use it in 2 different components, so might be moved to `helpers`

src/components/CartItem/CartItem.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+      case 'accessory':
+        return 'accessories';
+
+      default:
+        return '';
+    }
+  };
+
+  const handleClickDeleteItem = (productId: string) => () => {
```
TarasHoliuk, line 30:
```suggestion
  const handleDeleteItem = (productId: string) => () => {
```
and it's better to define these functions in the `CartPage` component and pass them as props. To achieve the single responsibility principle

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { HeaderNavLink } from './HeaderNavLink';
+import { HeaderMobileMenu } from './HeaderMobileMenu';
+import {
+  CartContext, FavoritesContext,
+} from '../../helpers/LocaleStorageContext';
+import { SearchInput } from '../SearchInput';
+import './Header.scss';
+
+export const Header = () => {
+  const navLinks = ['home', 'phones', 'tablets', 'accessories'];
```
TarasHoliuk, line 13: move it out of component

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const [hasMobileMenu, setHasMobileMenu] = useState(false);
+  const handleClickMenu = () => (setHasMobileMenu(true));
```
TarasHoliuk, line 15:
```suggestion
  const handleClickMenu = () => setHasMobileMenu(true);
```

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const { cart } = useContext(CartContext);
+  const { favorites } = useContext(FavoritesContext);
+  const location = useLocation();
+  const { productId } = useParams();
+  const hasSearchInput = (navLinks.some(link => (
+    location.pathname.includes(link) || location.pathname.includes('favorites')
```
TarasHoliuk, line 21: you can move `favorites` to array and rename it instead of use `||`

TarasHoliuk (changes requested):
You've done good work:+1: 
I've added a few comments to improve it

src/helpers/different.tsx
```diff
@@ -0,0 +1,19 @@
...
+export const transformType = (type: string) => {
+  switch (type) {
+    case 'phone':
+      return 'phones';
+
+    case 'tablet':
+      return 'tablets';
+
+    case 'accessory':
+      return 'accessories';
```
AlisaVlasova, line 14: it will be better to use enum type instead of this helper

AlisaVlasova (approved):

PR #159, opened 2023-03-31
Develop

ivanm2706:


ivanm2706:
[DEMO] https://ivanm2706.github.io/react_phone-catalog/

DarkMistyRoom (changes requested):
Great job. Let's fix some comments
- No need to store 0 amount of product in the cart. Remove it disable "-" when 1 is reached

image

- check these buttons. They start work not from first click
image

- 
image
image

- "Phones" is unclickable
image

- fix the placeholder
image

src/Components/Category/Category.tsx
```diff
@@ -0,0 +1,42 @@
+import { Link } from 'react-router-dom';
+
+export const Category = () => {
+  return (
```
etojeDenys, line 4:
```suggestion
export const Category = () => (
```

etojeDenys (approved):
lgtm

PR #176, opened 2023-06-06
Phone Book Solution

Zibi95:
[DEMO](https://zibi95.github.io/react_phone-catalog/#/)


DorotaLeniecDev (changes requested):
Please add demo and test report link :)

Zibi95:
I did only HomePage for now. Rest of tests don't pass so i don't get any link to see the result of tests.

src/styleUtils/_variables.scss
```diff
@@ -0,0 +1,7 @@
+$primary-text: #313237;
+$secondary-text: #89939a;
+$icons: #b4bdc4;
+$elements: #e2e6e9;
+$hover: #fafbfc;
+$green: #27ae60;
```
darokrk, line 6: good practice is to declare variables for colors like `$color-green`, `$color-red` etc.

don't use no descriptive names like `icons` or `elements` because they not pointing colors, rather something different.

if you have problems to name colors correctly you can try this [https://coolors.co/d1d1d1-dbdbdb-85c7f2-4a4855-4c4c4c](url) here you can paste you Hex values and it will generates for you helpful name :)

src/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,59 @@
...
+      </li>
+    </ul>
+
+    <ul className="nav__controls">
+      <li className="nav__item">
+        <NavLink title="Favorites" className={isLinkActive} to="/favorites">
+          <img
+            className="nav__icon"
+            src={favorite}
+            alt="Favorite icon button"
```
darokrk, line 48: This should be just `Favorites` don't need to add here more names

src/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,59 @@
...
+          />
+        </NavLink>
+      </li>
+      <li className="nav__item">
+        <NavLink title="Cart" className={isLinkActive} to="/cart">
+          <img className="nav__icon" src={cart} alt="Cart icon button" />
```
darokrk, line 54: Same here

src/components/ProductCardsContainer/ProductCardsContainer.tsx
```diff
@@ -0,0 +1,75 @@
...
+        <div className="cards-container__controls">
+          <button
+            className="cards-container__button"
+            type="button"
+            onClick={() => handleClick(-1)}
+            disabled={page === 0}
+          >
+            <img
+              src={leftArrow}
+              alt="Left arrow button to control the slider"
```
darokrk, line 51: Just `slider left arrow `

darokrk (changes requested):
Need some improvements

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,83 @@
...
+            style={{ transform: `translateX(${-100 * page}%)` }}
+          >
+            {bannerImages.map((image) => (
+              <li key={image}>
+                <img
+                  width={1040}
+                  height={400}
+                  className="banner__image"
+                  src={image}
+                  alt="Banner element"
```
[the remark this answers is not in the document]
Zibi95: I wanted to name it like that but eslint is fierce
image

mkyloa (changes requested):
check again and fix all the comments from previous reviews and send for the review again 👍

PR #181, opened 2023-06-13
react catalog

katerynashylina:
https://katerynashylina.github.io/react_phone-catalog


TarasHoliuk:
The home page link leads to a 404 page (on other pages as well):
image

Align content on the product page according to the design (also, it'd be nice for you to make sure it looks good on screens with 1920px width or more, not only 1440px. Most likely it will be open (by HRs) from laptops with such resolution):
image

src/components/Category/Category.tsx
```diff
@@ -0,0 +1,70 @@
...
+            <Link to="/phones">
+              <img src={phones} alt="phones" className="image" />
+            </Link>
+
+            <h2 className="image__title">
+              Mobile phones
+            </h2>
+
+            <p className="image__subtitle">
+              71 models
```
TarasHoliuk, line 29: it's better to use dynamic value

src/components/PhoneCard/PhoneCard.tsx
```diff
@@ -0,0 +1,103 @@
...
+            <div className="phone__description">
+              <div className="phone__description--names">
+                Screen
+                {' '}
+                <br />
+                {' '}
+                Capacity
+                {' '}
+                <br />
+                {' '}
```
TarasHoliuk, line 69: use string template

src/App.tsx
```diff
@@ -1,9 +1,188 @@
...
-    <h1>React Phone Catalog</h1>
-  </div>
-);
+const App = () => {
+  const [phones, setPhones] = useState<Phone[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [searchResults, setSearchResults] = useState(phones);
+  const [searchQuery, setSearchQuery] = useState('');
+  const [likedProducts, setLikedProducts] = useState<Phone[]>([]);
+  const [cartProducts, setCartProducts] = useState<CartItem[]>([]);
```
TarasHoliuk, line 26: I'd recommend you create `useLocalStorage` custom hook to handle values stored in the local storage instead of saving them here and passing them as props to child components
you may google it or ask in the chat if you need help

TarasHoliuk (changes requested):
Very good job! Looks amazing :fire: 
I added a few comments - check them

src/components/Category/Category.tsx
```diff
@@ -0,0 +1,77 @@
...
+            </Link>
+
+            <h2 className="image__title">
+              Mobile phones
+            </h2>
+
+            <p className="image__subtitle">
+              {phones.length}
+              {' '}
+              models
```
TarasHoliuk, line 36:
```suggestion
              {`${phones.length} models`}
```
fix all similar cases
the variant with string template looks much laconic

TarasHoliuk (approved):
Well done :+1: Added one comment - fix it (easy to do, I'm sure you will do it right - so approved)

PR #184, opened 2023-06-30
Develop

Zibi95:
[DEMO](https://zibi95.github.io/react_phone-catalog/#/phones/apple-iphone-7-32gb-rosegold)


DorotaLeniecDev (changes requested):
@Zibi95 I will skip this for this moment as I would like to first check smaller PR (after weekend we have plenty), and our plugin does not allow that

DorotaLeniecDev (changes requested):
The Application looks really neat! 
The code base is rally good.
All requirements are met. What about tablets and accessories? They are not exposed by the API?

One suggestion form my side is to make use of path aliases so imports are not that long and odd looking

package.json
```diff
@@ -34,20 +34,21 @@
...
     "node-sass": "^6.0.1",
     "prettier": "^2.8.8",
+    "react-app-rewired": "^2.2.1",
     "stylelint": "^13.13.1",
     "typescript": "^4.6.3"
   },
   "scripts": {
-    "start": "mate-scripts start -l",
-    "build": "mate-scripts build",
+    "start": "react-app-rewired start -l",
```
DorotaLeniecDev, line 42: Good job with setting up path aliases! 💪

DorotaLeniecDev (approved):
This is looking good! Great job 💪  Keep up!

PR #186, opened 2023-07-04
Develop

LBondarchuk:
https://lbondarchuk.github.io/react_phone-catalog/


agniya-i (changes requested):
Great job!  🔥
Let's fix some small issues

1. Image shouldn't look stretched on the homepage slider 
Screenshot 2023-07-04 at 14 09 56

2. On the product check the design one more time and align the product slider block and description block. Also, align prices and check the font on the Add to Cart button   
Screenshot 2023-07-04 at 14 13 16

3. On the checkout page there is also a problem with the button font 
Screenshot 2023-07-04 at 14 16 25

LBondarchuk:
fixed

Anastasiia-Svintsova (changes requested):
- Download the logo from the Figma. It should be an SVG image, not a text 
image

- Check the font style everywhere. There is incorrect font-family
image
image

- Fix the input UI according to Figma 
image

- The icons size doesn't match as well 
image
image

- Why does this page under development? It should have the same UI as the phones page
image

- Add an action to the click on the checkout button. Shoq the message that the feature will be implemented soon, for example 
image

- Let's imagine that there will be 100 pages 🤔 Change it to match the following structure `1 2 3 4 5 ... 18` or `1 ... 2 3 4 ... 18`. Or maybe your own solution
image

LBondarchuk:
done

sTorba24 (changes requested):
Good work! Lets improve it

1. It looks weird
image

2. Would be good to see message about not found item
image

3. when i put `space` in search no matches found
image

LBondarchuk:
As for the photo, I like how it looks

LBondarchuk:
Almost done)
My fonts are slightly different from figma (namely weight 600), I replaced with 500 so it seems more similar to me.

polosanya (approved):
Great! Well done!

PR #193, opened 2023-07-23
Develop

vasyliev-anton:
https://vasyliev-anton.github.io/react_phone-catalog/


DarkMistyRoom (changes requested):
Great job!
- please fix the placeholder
image

- lead to the top of the page on opening the section([VIDEO](https://www.loom.com/share/6b8080a2257841f080ebac6cbf39a5e3?sid=48b34e13-07e7-494b-b5f2-4eee370bd25f))

- remove `query=` when it's empty
image

Suggestions for improvement:
- it would be great to make sliders work smoothly

- and make the search order-insensitive
image

src/Reducer/store.ts
```diff
@@ -0,0 +1,35 @@
...
+
+// типы для действий и состояния
+export type RootState = ReturnType<typeof store.getState>;
+export type AppDispatch = typeof store.dispatch;
+export type AppThunk<ReturnType = void> = ThunkAction<
+ReturnType,
+RootState,
+unknown,
+Action<string>
+>;
```
DarkMistyRoom, line 31:
```suggestion
export type AppThunk<ReturnType = void> = ThunkAction<
  ReturnType,
  RootState,
  unknown,
  Action<string>
>;
```

DarkMistyRoom: can we avoid `unkown` here?

DarkMistyRoom (approved):
Great!
Sorry, I didn't mention it in the first review, but:

- make the search field like in the mockup(close button on the right, inputted text with the same padding as placeholder)
image

- It would be nice if the photo was a link, just like the title
image

src/Reducer/store.ts
vasyliev-anton: Ty :)

PR #199, opened 2023-08-04
added solution

sheva10barca:
[DEMO LINK](https://sheva10barca.github.io/react_phone-catalog/)


src/pages/TabletsPage/TabletsPage.tsx
```diff
@@ -0,0 +1,58 @@
...
+      const normalizedName = tablet.name.toLowerCase().trim();
+
+      return normalizedName.includes(normalizedQuery);
+    });
+  }, [tablets, query]);
+
+  return (
+    <div className="TabletsPage">
+      <div className="container">
+        {filteredTablets.length === 0 && query && (
```
sTorba24, line 35:
```suggestion
        {!filteredTablets.length && query && (
```

src/pages/TabletsPage/TabletsPage.tsx
```diff
@@ -0,0 +1,58 @@
...
+          <NoSearchResults />
+        )}
+
+        {filteredTablets.length === 0 && !query ? (
```
sTorba24, line 39:
```suggestion
        {!filteredTablets.length && !query ? (
```

sTorba24 (changes requested):
Good work!

1. for broken link just white page, add `not found` page please
image

VitaliyBondarenko1982 (changes requested):
Good job!
Consider some improvements.
- make responsive header and footer and fix horizontal scroll
Screenshot from 2023-08-05 16-00-24
Screenshot from 2023-08-05 16-04-44
- center content on phones page fot tablet and mobile width. check other pages
Screenshot from 2023-08-05 16-03-00

src/components/ProductCard/ProductCard.tsx
```diff
@@ -0,0 +1,184 @@
...
+  useEffect(() => {
+    setIsLoading(true);
+
+    const loadProduct = async () => {
+      try {
+        const productFromServer = await getProduct(product.id);
+
+        setProductWithDetails(productFromServer);
+      } catch {
+        throw new Error('Loading Error');
```
anastasiia-tilikina, line 53: do not throw error in `catch` block it'll cause uncaught error in your script. Do something instead. for instance, show some error message to user.

src/components/ProductsSlider/ProductsSlider.scss
```diff
@@ -0,0 +1,79 @@
...
+    margin-top: 24px;
+    gap: 16px;
+
+    @include onTablet {
+      justify-content: center;
+      flex-wrap: wrap;
+    }
+  }
+
+  // &__card {
```
anastasiia-tilikina, line 74: delete comments

src/components/Search/Search.tsx
```diff
@@ -0,0 +1,68 @@
...
+        value={appliedQuery}
+        onChange={(event) => handleQueryChange(event)}
+      />
+      <button className="Search__btn" type="button">
+        {query ? (
+          <img
+            src={crossIcon}
+            alt="close"
+            data-cy="searchDelete"
+            onClick={handleClearQuery}
```
anastasiia-tilikina, line 60: why should you put click event on img instead of button and disable linter?

anastasiia-tilikina (changes requested):
Check comments.

Also let's fix some issues:

1. when I click on link in mobile menu it should close and show chosen page.
2. when I click on suggested product on `ProductDetailPage` it should open this page and scroll to top
3. now your carousel on mob devises is a bit confusing) let's show products in one row and jus increase or decrease number of cards depending on screen width.
image
4. looks like your search input doesn't allow to type in upper case. This is not desired behaviour. input value should be shown as is. transforming to lower case should be only for comparing.
(OPTIONAL)
5. Also I would suggest you to look at your `TabletsPage` and `AccessoriesPage` looks like if there were more products on those pages, they wouldn't work as desired. I mean pagination and filters. They should be rendered conditionally depending on how many products we have.
6. Also If we look closer to product pages we'll see that they're sharing the same logic) the difference only in url param and array of products that we should render. So technically it could be the same component which accepts props. It's up to you wether you want refactor this part or not though)

src/App.tsx
```diff
@@ -75,7 +73,12 @@ const App: React.FC = () => {
                 <Route index element={<HomePage products={products} />} />
 
                 <Route path="phones">
-                  <Route index element={<PhonesPage phones={phones} />} />
+                  <Route
```
anastasiia-tilikina, line 76: Great you did it with your product page!
Let's improve a bit more)
let's make our paths to product pages as variables for instance `:category` and make nested route. props will be passed depending on url

anastasiia-tilikina (changes requested):
You're doing Great job! very well done!

check my comment.
Also please check your breakpoints for media queries
in some width your layout is collapsed

image


image

OPTIONAL
In my opinion I'd suggest you to add smooth appearance for mobile menu

anastasiia-tilikina (changes requested):
You did a really great job! 🔥

I was browsing through your app and found 2 more small bugs! Let's bring your website a bit closer to perfect)))

image

when I type some query on mobile device menu should also close once query is applied. the same with clearing it.

This link in footer leads to home page but doesn't scroll to top of the page and it looks a bit confusing. Let's add scroll to top.

image


I believe this is our last fix) everything else looks great!

VitaliyBondarenko1982 (approved):
Well done!

PR #209, opened 2023-08-24
solution

vpdrabynko:
[DEMO LINK](https://vpdrabynko.github.io/react_phone-catalog/)

### DIFF
--- a/src/pages/FavouritesPage/Favourites.tsx
+++ b/src/pages/FavouritesPage/Favourites.tsx
@@ -25,7 +25,7 @@ export const Favourites: React.FC<Props> = ({
 
         <div className="favourites__content">
           <h1 className="favourites__title">
-            Favourites
+            Favorites
           </h1>
 
           <p className="favourites__subtitle">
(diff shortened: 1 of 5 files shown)


lerastarynets (changes requested):
Good job overall, you need to move reusable parts of code to separate components
also just a question, is this blue outline on select focused an expected behaviour?
Screenshot at Aug 24 12-41-41

sTorba24 (changes requested):
Well done! Lets improve it

I changed link manually and it shouldn't find phone, but i still see page which i broke(should be `not found` page
image

a little tricky that clicking on `add to cart` removing element when it's been added 
image


there are 9 items, not 2 :(
image

TarasHoliuk:
typo:
image

should be Favorites

src/pages/FavouritesPage/Favourites.tsx
```diff
@@ -25,7 +25,7 @@ export const Favourites: React.FC<Props> = ({
 
         <div className="favourites__content">
           <h1 className="favourites__title">
```
TarasHoliuk, line 27: filename and all classes are still with typo
You may do this in VSCode (FYI):
image

TarasHoliuk (approved):
Left 1 comment - that's only a mechanic fix - no need to re-request review
So Approved :heavy_check_mark: :+1:

PR #212, opened 2023-08-28
Develop

Sergey-Mironenko:


etojeDenys (changes requested):
please, add the demo link to PR description

Sergey-Mironenko:
https://sergey-mironenko.github.io/react_phone-catalog/

etojeDenys (changes requested):
1. this button in Breadcrumb should lead to home page
image


2. should be the links to the pages with these goods
image



3. looks like these btns are not working
image

lerastarynets (changes requested):
Well done with fixing comments)
Screenshot at Aug 30 17-10-47
Double click on logo, user is redirected to #/home page and no content/error message is displayed on the page, pls fix

src/App.tsx
```diff
@@ -1,9 +1,223 @@
...
-    <h1>React Phone Catalog</h1>
-  </div>
-);
+const App = () => {
+  const { setQuery } = useContext(Context);
+  const { pathname } = useLocation();
+  const history = useNavigate();
+  const [searchParams, setSearchParams] = useSearchParams();
+  const [filterType, setFilterType] = useState('');
+  const page = searchParams.get('page' || '');
```
etojeDenys, line 27:
```suggestion
  const page = searchParams.get('page')  || '';
```

src/App.tsx
```diff
@@ -1,9 +1,223 @@
...
+  const perPage = searchParams.get('perPage' || '');
+  const sort = searchParams.get('sort' || '');
+  const filterQuery = searchParams.get('query' || '');
+  const activeProduct = JSON.parse(
+    localStorage.getItem('product') as string,
```
etojeDenys, line 32: better to move all these query param names to some enum and reuse them everywhere
(the same i suggest making with localStorage keys)

src/App.tsx
```diff
@@ -1,9 +1,223 @@
...
+          searchParams.append(key, String(part));
+        });
+      } else {
+        searchParams.set(key, value);
+      }
+
+      setSearchParams(searchParams);
+    });
+  };
+
```
etojeDenys, line 53: you can try to create your custom hook using useSearchParams and this function

src/App.tsx
```diff
@@ -1,9 +1,223 @@
...
+        filterQuery={filterQuery}
+        updateSearch={updateSearch}
+      />
+
+      <main style={{ flexGrow: 1 }}>
+        <Routes>
+          <Route
+            path="/"
+            element={(
+              <Home pathname={pathname} />
```
etojeDenys, line 85: why do we need to send pathname to each route, if we can easily get that pathname using some of react-router hook

src/App.tsx
```diff
@@ -1,9 +1,223 @@
...
+            element={(
+              <>
+                <Navigation
+                  pathname={pathname}
+                />
+                <Products
+                  productType={ProductTypes.accessories}
+                  pathname={pathname}
+                  type="accessory"
+                  updateSearch={updateSearch}
```
etojeDenys, line 169: when you create a custom hook, this prop will be redundant because the usage can simply use this hook in the Products component

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -1,9 +1,187 @@
...
+  const [searchParams] = useSearchParams();
+  const [filterType, setFilterType] = useState('');
+  const page = searchParams.get(SearchTypes.page) || '';
+  const perPage = searchParams.get(SearchTypes.perPage) || '';
+  const sort = searchParams.get(SearchTypes.sort) || '';
+  const filterQuery = searchParams.get(SearchTypes.query) || '';
+  const activeProduct = JSON.parse(
+    localStorage.getItem(LocaleStorageTypes.product) as string,
+  ) || null;
+  const updateSearch = useUpdateSearch();
```
etojeDenys, line 37: you can also return the searchParams value, to avoid redundant useSearchParams usage

src/Header.tsx
```diff
@@ -0,0 +1,158 @@
...
+    } else {
+      updateSearch({ query: null });
+    }
+  };
+
+  useMemo(() => {
+    if (filterQuery) {
+      setQuery(filterQuery);
+    }
+  }, [filterQuery]);
```
etojeDenys, line 41: better to move this condition to useEffect

src/ProductCard.tsx
```diff
@@ -0,0 +1,121 @@
...
+              <div className="product__info--value">{product.ram || '-'}</div>
+            </div>
+            <div className="product__button_container">
+              <button
+                type="button"
+                className={classNames(
+                  'product__button_add',
+                  {
+                    'product__button_add--active':
+                    findProductOnCart(product.id) === true,
```
etojeDenys, line 80: let's try to find the product somewhere in component body and save the result in var to avoid redundant function call

src/Products.tsx
```diff
@@ -0,0 +1,391 @@
...
+      updateSearch({
+        page: null,
+        perPage: null,
+        sort: null,
+      })
+    );
+  }, [pathname]);
+
+  useMemo(() => {
+    if (pathname) {
```
etojeDenys, line 115: the same here, better to move it to useEffect

src/ProductsSlider.tsx
```diff
@@ -0,0 +1,155 @@
...
+  title,
+  dicount,
+  random,
+}) => {
+  const [position, setPosition] = useState(0);
+  const [isLoading, setIsLoading] = useState(false);
+  const [isError, setIsError] = useState(false);
+  const [products, setProducts] = useState<Product[] | []>([]);
+  const step = (272 + 25) * 4;
+  const frameSize = (272 * 4) + (25 * 3);
```
etojeDenys, line 26: move all these magic numbers to some vars

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -1,9 +1,185 @@
...
+      setFilterType('phones');
+    } else if (pathname === '/tablets') {
+      setFilterType('tablets');
+    } else if (pathname === '/accessories') {
+      setFilterType('accessories');
+    } else if (pathname === '/favourites') {
+      setFilterType('favourites');
+    } else {
+      setFilterType('');
+    }
```
etojeDenys, line 48: move to useEffect 

try to read where we should use [useMemo](https://react.dev/reference/react/useMemo)

etojeDenys (changes requested):

src/ProductsSlider.tsx
```diff
@@ -22,10 +22,13 @@ export const ProductsSlider: React.FC<Props> = ({
   const [isLoading, setIsLoading] = useState(false);
   const [isError, setIsError] = useState(false);
   const [products, setProducts] = useState<Product[] | []>([]);
-  const step = (272 + 25) * 4;
-  const frameSize = (272 * 4) + (25 * 3);
+  const cardWidth = 275;
+  const gap = 25;
+  const cardsPerPage = 4;
+  const step = (cardWidth + gap) * cardsPerPage;
+  const frameSize = (cardWidth * cardsPerPage) + (gap * cardsPerPage - 1);
```
etojeDenys, line 29: looks like now we can move them out of the component as constants

etojeDenys (changes requested):

etojeDenys (approved):
looks good

PR #219, opened 2023-09-08
this project is completed(need check)

AndriiYelieva:
[DEMO LINK](https://AndriiYelieva.github.io/react_phone-catalog/)


VitaliyBondarenko1982 (changes requested):
.

VitaliyBodnar (changes requested):
- All tests are not passed 
Screenshot 2023-09-09 at 14 22 06

sTorba24 (changes requested):
Well done!


1. should be device name here and in breadcrumbs
image

2.now here are 5 items in cart
image
 
3. search isn't working 
image


4. add mobile and tablet version :)

volodymyr-soltys97 (changes requested):
1. Add transition for hover effects
image

2. Will be better if you add hover effects for all clickable elements
image

volodymyr-soltys97 (changes requested):
1. These links should be opened in the new tab
image

2. I think it would be better if the product disappears from the shopping cart when you click on this button
image

VitaliyBondarenko1982 (approved):
Looks good

PR #232, opened 2023-09-23
React Phone Catalog

dudar-y:
[DEMO LINK](https://dudar-y.github.io/react_phone-catalog/)


src/components/BackLink/BackLink.scss
```diff
@@ -0,0 +1,21 @@
+@import "../../utils/mixins";
+@import "../../utils/variables";
```
TarasHoliuk, line 2: it's better to import such things once in some main.scss

src/utils/_mixins.scss
```diff
@@ -0,0 +1,117 @@
+@import "./variables";
```
TarasHoliuk, line 1: these scss files should be inside styles folder, there (src/utils) i'd expect to see js functions utils

src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,29 @@
+import { useState } from 'react';
+import { Product } from '../types/Product';
+
+export function useLocalStorage(
+  key: string, defoultItems: Product[],
+): [Product[], (i: Product[]) => void] {
```
TarasHoliuk, line 6:
```suggestion
export function useLocalStorage<T>(
  key: string, defoultItems: T,
): [T, (i: T) => void] {
```
it's better to make it more reusable (use generic type) to allow saving anything, or rename the hook accordingly to show we save products only in the local storage

TarasHoliuk (changes requested):
Well done 🔥 
Let's improve a few things

src/ShopProvider.tsx
```diff
@@ -0,0 +1,106 @@
...
+      });
+  };
+
+  const addToFavourites = (id: string): void => {
+    findProduct(id)
+      .then(product => {
+        if (product) {
+          const newProduct = { ...product };
+
+          newProduct.count = 1;
```
etojeDenys, line 46:
```suggestion
          const newProduct = { ...product, count: 1 };
```

etojeDenys (approved):
lgtm

PR #235, opened 2023-09-29
Solution

olena-hapon:
https://olena-hapon.github.io/react_phone-catalog/


sTorba24 (approved):
Weel done! Lets improve it

1. make this transition a bit faster :) it's too slow)
image

2. red label with number means you have something
0 means you haven't, so show label only when items exists in favorites / cart
image

PR #249, opened 2023-10-22
Develop

Reject-4444:
https://reject-4444.github.io/react_phone-catalog/

According to the task i've it done almost (without main slaider and some pages). I've spent for it one week. The problem is that i've done it using old api (according to the task). Now i've founded out that i have to use '_new' but not 'api'. If i need change 'api' to '_new' i have to change almost logic in this code


Esceype (changes requested):
Good job, there are a few features and pages left to finish, continue with the data you have so far

etojeDenys (changes requested):
looks good, just adjust your layout for mobile devices

xiimiik (changes requested):
(u dont need to do mobile adaptive)

Current problems:

jumping on click
image

i changed sorter in accessories and moved to tablets tabs and got crash
image

on wide screen content should be centred
image

add favicon
image

etojeDenys (approved):
good job

PR #250, opened 2023-10-22
solution

Kinderwagen:
[DEMO LINK](https://Kinderwagen.github.io/react_phone-catalog/)


etojeDenys (changes requested):
1. let's make the whole button clickable
image


2. should be Added to to cart
image

volodymyr-soltys97 (changes requested):
1. When I click favorites I get this problem with line
image

2. If there are no pages, then the footer should be located evenly at the bottom of the screen 
image

3. Add a favicon on the page
image

4. Currently the input focuses when clicking on specific text only, try changing the logic so that the input focuses even when clicking on an empty area
image

loralevitska (approved):
Good job!
Check my comment for improvement 
image

PR #262, opened 2023-11-15
add task

VladyslavKolisnyk:
DEMO LINK: https://vladyslavkolisnyk.github.io/react_phone-catalog/


maxim2310 (changes requested):

1.button back working strange, if i go to selected phone page and click on some color or change memory and then click back i expected that i will go to phone catalog but i back to previous selected memory or color 
2. need to add debounce on search and dont make search if nothing change in search (example if i have "Apple iPhone 8" and delete and return 8)

anastasiia-tilikina (changes requested):
Well done! You did great job!

1. back button should go to the catalog, not to home (you can set to={'../'} path
2. when I change page in pagination this page should be shown at the top

xiimiik (approved):
nice work!

PR #269, opened 2023-11-24
part1

Marinakyrychynska:
DEMO LINK: https://Marinakyrychynska.github.io/react_phone-catalog/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon on the page
image

image

Keep going with other steps and logic for your code

lerastarynets (changes requested):
Pls watch the video review [via the link](https://www.awesomescreenshot.com/video/23246307?key=4cd174bf042f44033cc63ac73c511c44) :)

sTorba24 (changes requested):
Good work!

1. carT *(actually cart is only for `cart` page, for favourites better to use `no favourites items or smth similar`)
image

2. remove borders here or add some padding inside to not stick text to border
image


3. it's loading, it's not `no result` :) fix please

image

4. it's not best message related to UX, because generally you just don't have tablets available, so tell it like `no products on this section`

image

5. add some x-paddings for mobile version
image

etojeDenys (approved):
good job

PR #274, opened 2023-11-29
completed task

Horizon-git:
[DEMO](https://horizon-git.github.io/react_phone-catalog/#/tablets)


Horizon-git:
i fixed everything. But on the non-iphone API there is no colors and capacity, however color and capacity on Figma design, so I did it by myself, but it is not good cuz of I dont't have data from the backend.

TarasHoliuk (approved):
Looks amazing :fire: 
Good compromise with color and capacity selectors :+1: 

Generally, it was designed for new API (for products in `public/_new`)

PR #280, opened 2023-12-04
Dev

victorgri:
https://victorgri.github.io/react_phone-catalog/


sTorba24 (changes requested):
Good work! Generally looks cool!  Lets improve it

use grid here, to look it like
item item item item
item item
image

maxim2310 (approved):
Looks nice, just check this active style
image

PR #281, opened 2023-12-04
Develop

AngerDESTROYS:
-[DEMO](https://angerdestroys.github.io/react_phone-catalog/)


l4st1m0za (changes requested):
Great job, but some changes required:

1) This is shown as I open the website
image

2) It's better to add some hover effect on these links
image

3) Make this as in the design 
Your:
image
Design:
image

4) When the phone is missing, message is shown in the different place according to other messages
image
image

5) Correct the phrase in here
image

6) Heart has white outline, you can use .svg for reducing such things
image

7) You can also add some hover effect
image

8) Back button in the cart should work as the web button back, but insted I'm getting redirected to the home page

PR #283, opened 2023-12-09
start development

RomanOstrous:
https://RomanOstrous.github.io/react_phone-catalog


src/services/FethClient.ts
```diff
@@ -0,0 +1,47 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
TarasHoliuk, line 1: such files should be named in the PascalCase - use camelCase

TarasHoliuk (changes requested):
Demo doesn't work and the task isn't ready yet:

image

Please, re-request review when the task will be done
In case you have any questions or need help - feel free to ask in the chat or join to the Q&A session :)

PR #288, opened 2023-12-16
not ready

Kozubowicz:
Not yet ready


anastasiiavorobiova (changes requested):
Make sure to deploy your page

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add hover effects for all clickable elements, fox example for the nav menu on the header

2. Also need to add a transition for hover effects

anastasiiavorobiova (approved):
Awesome work!

PR #290, opened 2023-12-17
Develop

ydem1:
https://ydem1.github.io/react_phone-catalog/


l4st1m0za (changes requested):
Good job, but some changes needed:

1) Use .svg for this photos as now they are not of the best quality 
image

2) When I open "Phones" tab it states they phones are sorted by "Newest" but in this case it's false
image

But when I click on newest then everything is okay
image

sTorba24 (approved):
Nicely done!

PR #291, opened 2023-12-18
Develop

dimarogkov:
https://dimarogkov.github.io/react_phone-catalog/


IvanFesenko (changes requested):
Grate work. Check my comments:
- add fav icon
- Search should apply filtration on current page withoot redirecting
https://github.com/mate-academy/react_phone-catalog/assets/61904995/43fbb1b7-1e9f-4674-8b12-f6306eb3aecd
- pagination issue
https://github.com/mate-academy/react_phone-catalog/assets/61904995/da067e6f-2caa-4361-a979-3878811804c4

denys-danyliuk (changes requested):
text in input should be corresponding to the section user currently in
Знімок екрана 2023-12-20 о 14 07 22

add the dollar sign before price, check the design
Знімок екрана 2023-12-20 о 14 09 13

images are being cut out on hover, also try to make hover effect more visible
Знімок екрана 2023-12-20 о 14 10 17

when I choose some phone from this list would be better if page was scrolled to the top because now it's unclear that page is changed
Знімок екрана 2023-12-20 о 14 11 42

also when filtering phones using input use debounce 

make this text as a separate element, now it doesn't look good, you can show it for 3 seconds and then hide it


also name of the phone should be a link not just plain text (inside cart)
Знімок екрана 2023-12-20 о 14 14 43

SanyaBratashchuk (changes requested):
1. 1 item but 2 items 
image
2. there are price for 4 items 
image

maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

1. would be nice to add debounce on search 
2. when i select phone then change color and try to click back it must lead to previous page but now its just return to previous color
3. also would be nice to add feature went i click minus in and item has one quantity then this item will be removed from basket

DarkMistyRoom (changes requested):
Haven't you forgotten to push the commit? I see no commits since the last review

dimarogkov:
1 - this is not a mandatory task
2 - previous color - this is the previous page
3 - I didn't understand the essence of the problem, why delete to minus, if there is a cross icon, when you click on it, the product is removed from the cart.

IvanFesenko (changes requested):
Any updates with those comments? I dont see any commits as well. If you have questions - ask them in fe_chat
2023-12-21_15h39_39

IvanFesenko (changes requested):
Fix previous mentors comments

ivbutsykin (approved):
Great work!

PR #299, opened 2023-12-22
add task solution

Vovchukkul:
[DEMO LINK](http://localhost:3000/#/)


ivbutsykin (changes requested):
Please, fix some issues:
1. Your footer is fixed so it's sometimes can overlap content
Screenshot 2023-12-22 at 12 36 33

2. Show badges with count of items in favourites and basket
Screenshot 2023-12-22 at 12 37 40

3. Fix UI of not implemented feature:
Screenshot 2023-12-22 at 12 39 26

4. After reload items from basket disappear, you local storage to save them (similar as favourites).

sTorba24 (changes requested):
Nicely done!

1. fix images view, they are flattened
image

2. try to change device color, layout near image jumping(some px), fix it

3. check item detail page, a lot of small differences
image


now: 
image

4. should be gap between these buttons 
image


5. `-` button shouldn't work if it is only one item of this type.

6. also when i do it, then go `home` and trying to add this element again - it doesn't work, need to click twice

7. still not fixed view of `not implemented feature`
image

SanyaBratashchuk (changes requested):
1. It should be 1 item but many items 
image
2. Implement search

SanyaBratashchuk (changes requested):
There should be search 
image

Vovchukkul:
> There should be search image

but I`ve already implement it
image

SanyaBratashchuk (approved):

PR #301, opened 2023-12-25
solution

NikitaOrliak:
[DEMO LINK](https://nikitaorliak.github.io/react_phone-catalog/)


maxim2310 (changes requested):

1. would be nice to add hover on button 
image
2. when i click on plus, minus button is moving 
image
3. if i select phone > choose color > click back i expected that i return to phone page but i return trevious color
4. when select phone need to scroll up 
5. need to add debounce on search because now its just delay

etojeDenys (approved):

etojeDenys (changes requested):
still can reproduce it
image

sTorba24 (approved):
Nicely done!

PR #305, opened 2023-12-27
Develop

daryna-hnidash:
https://daryna-hnidash.github.io/react_phone-catalog/


sTorba24 (changes requested):
Good work!

1.Stretch header and footer for full page width
image


2. after input on search it redirect on home page

3. there are 5 items, not 2
image

DarkMistyRoom (changes requested):
Good job! Just make the tabs full-height clickable
image

src/components/ProductCard/ProductCard.scss
```diff
@@ -16,7 +16,7 @@
 
   &__img-container {
     display: flex;
-    justify-self: center;
+    margin: 0 auto;
```
TarasHoliuk, line 19: it seems like it doesn't work :cry:

TarasHoliuk:
Before adding to the portfolio:

1. Fix these links (they should not lead to 404 page):
image

2. Add the favicon

TarasHoliuk (approved):
Hey! Everything looks great, added 2 things to improve before adding it to the portfolio (small issues, so approved) :+1: 

And of course, if you will be able to fix the iPhone issue. But it's not critical, IMO :slightly_smiling_face: .

src/components/ProductCard/ProductCard.scss
daryna-hnidash: Oh... I`ve just seen it.. But I figured it out but my own)) justify-self seems the most suspicious to me))

PR #306, opened 2023-12-28
init

MaksymKos:
https://maksymkos.github.io/react_phone-catalog/


DarkMistyRoom (changes requested):
- check if you added all needed dependencies to package.json and specify all types https://github.com/mate-academy/react_phone-catalog/actions/runs/7345470071/job/19998673904?pr=306
- make this button lead to the top
image

- center the content
image

- would be good to use 4 columns in a row and limit the content width

- it shows all items when I choose 16 option
image

- these buttons seems don't work on the product details page
image

- these buttons should show a number of items. 
image
- use underline for hover effect
image

- + button doesn't work
image

- add hover effect and show the message that this functionality isn't implemented yet on click
image

TarasHoliuk:
1. The first point from the previous review is still not fixed (completely):
image

Also, make sure that @types version matches the library version (react-router-dom). Also, it's better to write all @types in devDependencies:
image

2. It was 2 different things)
image

it means add some hover effect and show message (below) that this functionality isn't implemented yet **_on click_**

3. On hover the border-bottom disappears:
image

4. Add the favicon

TarasHoliuk (changes requested):
GJ :+1: 
A few previous comments weren't properly fixed, also added a few new comments

IvanFesenko (changes requested):
If there are no results do not show pagination and page settings -you should show message like 'There are no results on ypour request' or so
2024-01-03_16h13_20

Viktor-Kost (changes requested):
### Great job!

### Noticed to fix:
1. add a favicon
2. as previous mentor said
`would be good to use 4 columns in a row and limit the content width`
So make 4 items in a row and center content. I see you did it at phones page so make the same at the main page (and product page - You may also like)
3. add hover effect to `Add to cart` button
4. hover effect shouldn't be separated
https://github.com/mate-academy/react_phone-catalog/assets/117587463/bc032ade-e814-4974-a3f6-6ea1cfaae777


### Recommendations:
1. would be nice if you tried to make this item fill the full available space. 
And as for me (recommendation) items should change on click faster than auto changing

2. when I go to a product from a favorites and then click `back` I should be lead to favorites, not to products
https://github.com/mate-academy/react_phone-catalog/assets/117587463/ef0b73b7-66da-429d-b23f-ed193667f675

IvanFesenko (approved):
Well done

PR #325, opened 2024-01-15
Portfolio project

VKdrvtsv:
[DEMO LINK](https://VKdrvtsv.github.io/react_phone-catalog/).


maxim2310 (changes requested):

1. img under button 
image
2. product img broken 
image
3. would be nice to have 4 in row, on phones page same
image
4. back button dont work if i change color 
image
5. after click on breadcrumbs search is not showing 
image
image
6. would be nice to add debounce on search
7. add padding and align items 
image

VKdrvtsv:
I make media-adoption also, so I would be thankfull if you check it.

SanyaBratashchuk (changes requested):
Try to rename _new folder to `new` then adjust your paths and redeploy page. It may solve issue with photos.

VKdrvtsv:
Looks like the site wokrs on the right way, thank you)  Please check also my adoptation for devices)

etojeDenys (approved):
lgtm

PR #330, opened 2024-01-25
add homePage

AntonBelia:
[DEMO LINK](https://AntonBelia.github.io/react_phone-catalog/)


TarasHoliuk:
1. Have you tried to fix it? Add the package.json file to git as well:
image

2. Github, contacts, and rights links should be centered. GitHub link should lead to your GitHub profile (external link - open in the new tab).
image

3. I see you've implemented the home page only (at least it's the only page with content in the deployed variant). But there is a problem with images:
image

4. I'd recommend you to use new products, i.e. this link to get them: https://mate-academy.github.io/react_phone-catalog/_new/products.json. (replace `api` in path to `_new` - the same as the original file structure in the project). Ask more in chat if it's not clear which links you should use to get new products

TarasHoliuk (changes requested):
Implement all pages, fix images on deployed version
For more info and help ask in the chat :slightly_smiling_face:

vadiimvooo (changes requested):
Good job. Here are a few improvements:
1. Fix the elements of the block to be the same height
image
2. Border should not dissapear on hover of card
image
3. Add transition effect for all interactive elements for your webpage (links in footer, header etc.)
4. Arrows in select elements are placed wrong.
image

DarkMistyRoom (approved):
Good job! Only one suggestion
- add hover effects to these links
image

PR #332, opened 2024-01-25
Solution

andreysakhno:
[DEMO LINK](https://andreysakhno.github.io/react_phone-catalog/)


src/libs/components/back/Back.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { IconNames } from '../../enums';
+
+type Props = {
+  className?: string;
+};
+
+export const Back: React.FC<Props> = ({ className }) => {
+  const { state, pathname } = useLocation();
+
+  const backPath = state ? `${state.pathname}${state.search || ''}` : '..';
```
ivbutsykin, line 16: **Perhaps** you can use simple button (instead Link) and on `onClick` use `navigate(-1)`: https://reactrouter.com/en/main/hooks/use-navigate

ivbutsykin (changes requested):
Good work! I found only two issues:
1. Looks like "Newest" filter sort in opposite side.
Screenshot 2024-01-26 at 11 28 29
2. "Back" button on phones screen should navigate like `navigate(ROUTES.PHONES)`. Because now you have a bug when user will choose different colours and then click "back" (expected result - redirect on phones page). 
Screenshot 2024-01-26 at 11 33 20

src/libs/components/back/Back.tsx
andreysakhno: I wanted the user to be able to return to /phones with saved search params (page, sort, etc.), and also the user to be able to go to the main page if he came from another domain, so I used the state. 
Redirecting on phones page when user will choose different colours and then click "back" makes sense.

IvanFesenko (changes requested):
Looks better, but therre are still some issues
- Add favIcon
- Only first image on slider is visible
https://github.com/mate-academy/react_phone-catalog/assets/61904995/a1fe61d5-10b8-4131-8256-fecea2727fae

l4st1m0za (approved):
Great job!

PR #333, opened 2024-01-29
React phone catalog

Shostyy:
[DEMO LINK](https://shostyy.github.io/react_phone-catalog/)


sTorba24 (changes requested):
Good work! Lets improve it

1.add cursor pointer to all interactive elements
image

2. Make whole card as a link to details (not title only)
image

3. How to show cart again? Show message only as an alert etc(without hidding other elements)

4. sometime scroll disappears
image

5. looks broken a little on mobile
image

anastasiia-tilikina (changes requested):
1. Pls make ALL interactive on your site to have cursor pointer (optionally but preferable - hover effects) Now some elements are missing this. For instance:

image
image
image

and maybe some other.

2. Problem with scrolling is still not fixed
3. There is some bug, I cannot increment or decrement items amount on this element
image

4. When I go to product details page while details are loading "Page not found" is shown. I believe there should be a loader instead

5. I would suggest you to center this block on mobile

maxim2310 (changes requested):
this still not fixed, check it plz and fix
image

l4st1m0za (approved):
Good job!

PR #337, opened 2024-02-13
My Phone Catalog

SashaViun:
https://SashaViun.github.io/react_phone_catalog/


TarasHoliuk (changes requested):
The demo link doesn't work :cry: 

image

Please, make sure your solution properly deployed (ask in the chat for help if you need) and re-request review

GUSILLUS (changes requested):
GJ! But:

1. Try to fix favicon
image

2. Add some hover effect for card item
image

3.  Try to reverse this effect dark by default original image on hover
image

4. Fix the search text style like on the design
image

5. Try to make custom dropdown for perPage select like on a design
image

PR #338, opened 2024-02-14
add task solution

netzelnatalia:
[DEMO LINK](https://netzelnatalia.github.io/react_phone-catalog/)


src/App.tsx
```diff
@@ -1,7 +1,23 @@
...
+import { Outlet } from 'react-router-dom';
+import { Footer } from './components/Footer';
+import { Header } from './components/Header';
+import { GeneralProvider } from './helpers/GeneralContext';
 
 export const App = () => (
-  <div className="App">
-    <h1>React Phone Catalog</h1>
-  </div>
+  <GeneralProvider>
```
yevhenii-pyl, line 8: Don't you find the name 'GeneralProvider' like... too general? :) 
It's not a mistake, but why don't we think about some alternative naming?

src/api/api.ts
```diff
@@ -0,0 +1,57 @@
+import { Product } from '../types/Product';
+
+export const BASE_URL
```
yevhenii-pyl, line 3: Nice. Creating a const for a basic url is a great approach.

src/api/api.ts
```diff
@@ -0,0 +1,57 @@
...
+const BASE_API_URL = `${BASE_URL}/_new/products`;
+
+export const getAllProducts = (): Promise<Product[]> => {
+  return fetch(`${BASE_API_URL}.json`)
+    .then(response => response.json());
+};
+
+export const getHotPriceProducts = () => {
+  return getAllProducts()
+    .then(products => products
```
yevhenii-pyl, line 15: Well, basically, you fill never do filtering like this on a FE side and this will always be handled on the backend. 
However, as a simplification for training purposes this may work.

src/api/api.ts
```diff
@@ -0,0 +1,57 @@
+import { Product } from '../types/Product';
```
yevhenii-pyl, line 1: I'd probably called this file `index.ts` and use it as a main entry point for all API calls. 

Then, create for each category: phones, tablets, all products just to have it nicely stored. 

Definitely, not a mistake here on your side as for current assignment, but as the app grows having all API calls in one file may become a bit of a pain.

src/components/CartItem/CartItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+    changedProduct.quantity += change;
+
+    const newCart = [...cartList];
+
+    newCart.splice(index, 1, changedProduct);
+
+    setCartList(newCart);
+  };
+
+  const increaseNumber = () => changeNumber(1);
```
yevhenii-pyl, line 38: Why don't you use changeNumber as a handler directly? 
Having that extra functions looks unnecessary.

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,85 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-disable global-require */
+import React, { useState } from 'react';
```
yevhenii-pyl, line 3: I'd change imports order for consistency: 
```
// external libs
import React, { useState } from 'react';
import { Link } from 'react-router-dom';

// components
import { Checkout } from '../Checkout';

// styles
import './Footer.scss';
```

src/index.tsx
```diff
@@ -1,5 +1,65 @@
...
+/* eslint-disable global-require */
 import { createRoot } from 'react-dom/client';
+import { Helmet } from 'react-helmet';
+
+import {
+  HashRouter as Router,
+  Routes,
+  Route,
+  Navigate,
+} from 'react-router-dom';
```
yevhenii-pyl, line 11: Can't see react-router-dom at the package.json

src/pages/PhonesPage/PhonesPage.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { GeneralContext } from '../../helpers/GeneralContext';
+import { getPhones } from '../../api/api';
+
+export const PhonesPage: React.FC = () => {
+  const { setIsLoading } = useContext(GeneralContext);
+  const [phonesList, setPhonesList] = useState<Product[]>([]);
+
+  useEffect(() => {
+    setIsLoading(true);
+    getPhones().then(setPhonesList)
```
yevhenii-pyl, line 14: What if there is a fetching error?

src/pages/ProductDetailsPage/ProductDetailsPage.tsx
```diff
@@ -0,0 +1,140 @@
...
+  let category = '';
+
+  if (productFromList) {
+    category = productFromList?.category.charAt(0).toUpperCase()
+    + productFromList?.category.slice(1);
+  }
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
```
yevhenii-pyl, line 33: catch?

src/helpers/GeneralContext.tsx
```diff
@@ -0,0 +1,52 @@
+import React, { useEffect, useState } from 'react';
```
yevhenii-pyl, line 1: It's not a helper, it is a context. I believe, that should be a separate folder for it like `store` or something.

yevhenii-pyl (changes requested):
Solid work!
Let's just go through failed tests and provide a few minor adjustments.

PR #345, opened 2024-02-24
Dev

HYPER2307:
[Demo link](https://hyper2307.github.io/react_phone-catalog/)


TarasHoliuk:
1. These links lead to 404 page:

image

Add the link to your GitHub as well, or scroll to the top, etc. 404 is definitely not the best behavior.

2. Make the header visible when I scroll down (i.e. sticky or fixed):

[Screencast from 25-02-24 11:01:30.webm](https://github.com/mate-academy/react_phone-catalog/assets/86923089/bf697f8c-c486-4bca-bcaa-6ee91795c7d5)

3. Checkout leads to `Not found page` and redirects. Just show a message :
image
image

4. Add a favicon

src/storage/phonesContext.tsx
```diff
@@ -0,0 +1,245 @@
...
+  preparedHotPriceProducts: Product[],
+  sortType: SortType,
+  setSortType: Dispatch<SetStateAction<SortType>>,
+  itemsPerPage: number,
+  setItemsPerPage: Dispatch<SetStateAction<number>>,
+  sortParams: SortParamsType[],
+  perPageParams: number[],
+  sortedProducts: () => Product[],
+  tabletSearchValue: string,
+  setTabletSearchValue: Dispatch<SetStateAction<string>>,
```
TarasHoliuk, line 25: it's better to use search params to store `searchValue`
No need to save it when you change the tab

same for all sort params

src/storage/phonesContext.tsx
```diff
@@ -0,0 +1,245 @@
...
+  phoneSearchValue: string,
+  setPhoneSearchValue: Dispatch<SetStateAction<string>>,
+  filteredProducts: Product[],
+  suggestedProducts: Product[],
+  setSuggestedProducts: Dispatch<SetStateAction<Product[]>>,
+  favoritesId: string[],
+  setFavoritesId: (v: string[]) => void,
```
TarasHoliuk, line 32:
```suggestion
  favoritesIds: string[],
  setFavoritesIds: (v: string[]) => void,
```

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,100 @@
...
+      return i - 1;
+    });
+  };
+
+  return (
+    <div className="pagination">
+      <div className="pagination__list">
+        {products.map(phone => (
+          <Fragment key={getId()}>
+            <ProductItem
```
TarasHoliuk, line 56: It's unclear from `Pagination` component that it contains the list of products.

`Pagination` component should contain only pagination:
image

src/storage/phonesContext.tsx
```diff
@@ -0,0 +1,245 @@
...
+
+  const preparedHotPriceProducts = products?.filter((product: Product) => (
+    product.fullPrice - product.price >= 90
+  )) || [];
+
+  const preparedBrandNewProducts = products?.filter((product: Product) => (
+    product.year >= 2019
+  )) || [];
+
+  const sortedProducts = () => {
```
TarasHoliuk, line 117:
```suggestion
  const getSortedProducts = () => {
```

src/storage/phonesContext.tsx
```diff
@@ -0,0 +1,245 @@
...
+      case SortType.Cheapest:
+        return [...products]
+          .sort((prev, next) => prev.fullPrice - next.fullPrice);
+
+      default:
+        return [...products].sort((prev, next) => next.year - prev.year);
+    }
+  };
+
+  const filteredProducts = sortedProducts().filter(product => (
```
TarasHoliuk, line 133: you have all sort and filter params available on every page. So it's better to recalculate it on the page where you use it (i.e. `PhonesPage`) using params

When it is in context you recalculate it on every page when any state value changes

TarasHoliuk (changes requested):
GJ :+1: 
Let's add a few improvements to your app and to code :slightly_smiling_face:

SanyaBratashchuk (approved):

PR #348, opened 2024-02-26
Solution for react_phone-catalog

petrovskyi-yevhen:
[DEMO LINK](https://petrovskyi-yevhen.github.io/react_phone-catalog)


DarkMistyRoom (changes requested):
- Would be great to sync the repo with the original to see the tests' results
- What if I want to sort the search results or set 4 items per page?
image

PR #351, opened 2024-02-29
Develop

atrofymovych:
https://atrofymovych.github.io/react_phone-catalog/


GUSILLUS (changes requested):
GJ! But:

1. Your site is crashing when I write something into search input on tables or accessories pages

2. I think it's better to make a spinner of products to show the next 4 items instead of next item

3. Add some hover effects to the item card
image

4. Same for these cards
image

5. Some UI bug when clicking on an add to cart button
https://www.loom.com/share/31e9b56cbe48421ba6fdf6fb04e17d31

6. Onlu red squares content need loaders or skeleton others need to be static
image

7.  Any change broke your website, pagination also
image

8. Need some notification logic
image

PR #352, opened 2024-03-02
Develop

Anmonnn:
[DEMO LINK](https://Anmonnn.github.io/react_phone-catalog/)


src/index.tsx
```diff
@@ -1,5 +1,51 @@
...
+import { Favourites } from "./pages/Favourites/Favourites";
+import { persistor, store } from "./store";
+import { Cart } from "./pages/Cart/Cart";
+import { ItemCard } from "./pages/ItemCard/ItemCard";
 
-createRoot(document.getElementById('root') as HTMLElement)
-  .render(<App />);
+createRoot(document.getElementById("root") as HTMLElement).render(
+  <Provider store={store}>
+    <BrowserRouter>
```
vadiimvooo, line 21: For gh-pages you need to use hash router instead.

vadiimvooo (changes requested):
Because of usage wrong routing your page seems broken, also there is no photos for carousel, arrows, icons etc.
image

sTorba24 (changes requested):
Good work!

1. 5 items in cart, but in summary 1 can see
image


2. there is discount for this phone 
image


3. fix icons 
image


4. should be vice versa sorting by cheapest
image

5. pagination isn't working
image


6. it is the same as mobiles but no items, so use the same layout with message `no items`
image


7. check detail page
image

GUSILLUS (changes requested):
GJ! But:

1. Why after clicking on a main slider after some click I redirecting to another page?
image

2. As a user I want to see the correct information
image
image
image

3. Also if you don't have an items on a page you don't need to show sorting and per page dropdowns

4. Better to scroll by 4 items per time because you have a lot of products
image

5. Add some hover effect on a phone card

6. Fix this
image

7. Try to hide this button if the page height equals window height

anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2024-03-17 at 16 39 57
The search vs sorting vs pagination don't work properly

PR #359, opened 2024-03-11
Solution

IrynaK-a:
Demo: https://irynak-a.github.io/react_phone-catalog/


IvanFesenko (approved):
Looks good. Greate job. I'd recomend use new data from api, cuz those phones looks like dinosaurs

IvanFesenko (commented):
Also pay attention that designs was updated https://github.com/mate-academy/react_phone-catalog and you should update if for portfolio

PR #364, opened 2024-03-19
Develop

vadymboichenkoAcupower:
https://vadymboichenkoacupower.github.io/react_phone-catalog/

There is a problem with CartPage useEffect method. Can not understand what to add to dependency to correct refresh a page


SanyaBratashchuk (changes requested):
1. missing dollar sign 
image
2. images is not displayed 
image

3. add favicon
4. set 1-st page as selected on change of items on page select 
5. add paddings 
image
6. font is not loaded

TarasHoliuk:
1. Images are squeezed (1st how it looks now - second how it should look. Use object-fit):
image

Also in the slider:
image

2. Cart counter is not updating after deleting product from the card:
[Screencast from 21-03-24 20:05:59.webm](https://github.com/mate-academy/react_phone-catalog/assets/86923089/e2025651-1c1f-48c7-b6a3-3c1a28433105)

Also, not sure it's a good idea to delete all products from the cart on `Checkout` button click - we show the message that this function is not implemented yet.

3. In such case display your text in the center of the page:
image

Also, even on the empty page there is a scrollbar. Do not use `overflow: scroll`, use `overflow: auto` instead.

4. Seems like footer has too big height:
image

TarasHoliuk (changes requested):
A few more fixes have to be fixed/improved.

Also, try to make the page a bit more adaptive. At least make sure there is no horizontal scroll and all content is displayed on mobile version. It wasn't necessary when you started doing it, but anyway, it will be part of your portfolio. So you can check new layout for different screens and implement it at least partially: https://github.com/mate-academy/react_phone-catalog

vadiimvooo (approved):
Well done!)

PR #365, opened 2024-03-21
Phone catalog

PodvAx:
[DEMO LINK](https://podvax.github.io/react_phone-catalog/)


loralevitska (changes requested):
You did an amazing job! Looks good! Let's do some improvements (check comments)

- github link open in new tab 
- open phone info by click on whole card not only on link
- add loader after phone request (filtering by search), for slow3g setting it works bad now

If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

vadiimvooo (approved):
Well done!)

PR #367, opened 2024-03-25
Develop

andrii-rykal:
[DEMO LINK](https://andrii-rykal.github.io/react_phone-catalog/)


IvanFesenko (changes requested):
Greate job. There are few things that should be fixed:
- Add favIcon
- in case swith phone by color `back` button should lead to phones page, not previous color model
2024-03-25_15h16_22
- Move search field according design
2024-03-25_15h17_25
- Add empty state message
2024-03-25_15h20_55
- Broken icons
2024-03-25_15h21_19
- It's price per item, not sum
2024-03-25_15h22_58

Viktor-Kost (changes requested):
### Great job!

### To improve:
1. still need to add a favicon (as previous mentor asked)


2. add hover effects to interactive elements
Screenshot 2024-03-25 at 18 30 40
Screenshot 2024-03-25 at 18 32 17


3. remove this block when cart is empty
Screenshot 2024-03-25 at 18 35 14


4. As we can see at task requirements there is should be a `modal dialog`. Open this modal at `Checkout` button click.
Screenshot 2024-03-25 at 18 37 26




### Recommendations:
1. We have no tablets and accessories at our catalog. 
Screenshot 2024-03-25 at 18 48 05

So here we can change text to something like this. Since we're not looking for this tablets we just don't have these tablets
Screenshot 2024-03-25 at 18 29 09

3. It's better to use a default value here like `Newest` instead of `Not sorted`
Screenshot 2024-03-25 at 18 45 55

DarkMistyRoom (changes requested):
Great job! Just a few more suggestions:
- Would be better if these links opened a new tab
image

- When I click "Add to favorite" button on the Product Details page it leads me to the top of the page. It is unexpected behavior. Please fix it
image

IvanFesenko (approved):
Well done

PR #375, opened 2024-03-31
Phone catalog

AllaSerhiienko:
DemoLink [https://allaserhiienko.github.io/react_phone-catalog/]


vadiimvooo (changes requested):
Good job, but there is one thing to fix:
1. Add possibility to remove from cart the phone which is already there by clicking to the button
image

AllaSerhiienko:
> Good job, but there is one thing to fix:
> 
> 1. Add possibility to remove from cart the phone which is already there by clicking to the button
>    image

Now I've done it, but the task condition says the following:
**If the product is already in the Cart the button should say Added to cart and do nothing.**

anastasiiavorobiova (approved):
Great work!

PR #382, opened 2024-04-14
Solution

denedweb:
https://denedweb.github.io/react_phone-catalog/

The design does not provide a search field.
I implemented it at my discretion as much as possible in accordance with the design.


volodymyr-soltys97 (changes requested):
Great work 🔥  
Let's improve your code
1. Add a favicon on the page
image

2. You need to update count on the icon
image
image

3. It will be better if you remove the product from the basket when you click again 
image

4. It's good that you implemented the search at your discretion, but the search should be here by design
image

DarkMistyRoom (approved):
Great job! Only one suggestion:
- it's optional but would be great to add smooth scroll to the slider
image

PR #397, opened 2024-04-24
Phone_catalog

alionazherdetska:
[DEMO LINK](https://alionazherdetska.github.io/react_phone-catalog/)


l4st1m0za (changes requested):
Demo link is not working

DarkMistyRoom (changes requested):
Good job! Here are some comments:
- limit the content width
image


- show an amount of added items to the favorites and to the cart
image

- check the structure of this section
image
image

- the indent here should be bigger
image

- make the cursor a pointer over these buttons
image

- as well as for these ones
image

PR #399, opened 2024-04-27
t

Ihor-Prodan:
https://ihor-prodan.github.io/react_phone-catalog/#/phones


l4st1m0za (changes requested):
Good job, but some changes needed:

0) +25 point for cool loader

1) This is an initial page of the app, it looks wrong
image

2) This logo is shrinked. Also it should be clickable
image

3) Your carousel is going a bit too fast, and this photo is not looking good
image

4) This element should have something
image

5) This link is not working 
image

6) No elements are loading on the page. Also, somehow the pagination is set to 46, and if i click on another page, i get teleported to the home page  
image

7) These links are not working, also
image

l4st1m0za (changes requested):
1) Use svg for better quality of the icons
image

2) Check info about the sales, as now you offer bigger price than sale
image

3) Quantity added to a specific item, should get counted in overal counter
image

l4st1m0za (changes requested):
image

Here is still no changes. See that in the first item, i have added 100 of them, and in the counter we still have 3 items only, but the price is increased accordingly

l4st1m0za (approved):
Good job!

anastasiiavorobiova (changes requested):
Great work! Check the layout here 
Screenshot 2024-05-14 at 15 48 09
Screenshot 2024-05-14 at 15 48 43
Consider fixing the counter in the cart 
Screenshot 2024-05-14 at 15 47 14

Ihor-Prodan:
Перед перевіркою завдання почистіть будь-ласка  local storege, інакше демо не відкриється через зміни )

GUSILLUS (changes requested):
GJ! 

1. Consider increase step of scrolling to 4 per time
image

2. Add some transition duration on all hover effects with the same time 

3. Remove this rainbow effect 
image

4. Fix the UI of back to top button. It always need to be the same to improve UX
image
image

4. Fix footer UI, height need to be fixed on every page
image

5. Some bug with dark theme
image

6. You can remove scroll here 
image

7. Add some logic on the checkout button
image

8. Made items card similar
image

9. Same with scroll
image

10. Some bug with dark theme 
image

11. Same
image

vadiimvooo (changes requested):
Good job. A few fixes required.
`Add some transition duration on all hover effects with the same time`
Content of this page should be centered. You have to use max-width for productDetail container class

GUSILLUS (changes requested):
GJ!

1. Let's rename to remove from cart
image

2. Footer bug
image

3. Fix the image size
image

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. Add cursor pointer for this button and transition for hover effects here 
image

2. The height for these links should be 48px 
image

3. You need to add transition when this menu appears on the page 
image

SanyaBratashchuk (approved):
apply `cursor: pointer` for this button
image

PR #401, opened 2024-04-30
Develop

kostyaniekrasov:
[DEMO LINK](https://kostyaniekrasov.github.io/react_phone-catalog/)


volodymyr-soltys97 (changes requested):
Great work 🔥 
The switch theme is also great!
Let's improve something
Currently there is no option to select more than 10 units, also the quantity does not update after we increase or decrease the items
image
image
image

When you click add again, it will be better to remove the product from the cart
image

GUSILLUS (changes requested):
GJ! 

1. Some bug with translation
image

2. Create a logic to show this button only if you have scroll
image

3. Disable active lang
image

lerastarynets (approved):
You're showing incredible skills, keep it going!

PR #407, opened 2024-05-13
Finish categories section for mobile

olafchuszno:
[DEMO LINK] https://olafchuszno.github.io/react_phone-catalog/


src/helpers/getProducts.ts
```diff
@@ -0,0 +1,12 @@
+import { Product } from '../types/Product';
+import { ProductCategories } from '../types/ProductCategories';
+
```
yevhenii-pyl, line 3: Let's create a base url:
export const BASE_URL = 'https://olafchuszno.github.io/react_phone-catalog/';

src/helpers/getProducts.ts
```diff
@@ -0,0 +1,12 @@
...
+export const getProductsByCategory: (
+  category: ProductCategories,
+) => Promise<Product[]> = (category: ProductCategories) => {
+  return fetch(`/api/${category}.json`)
```
yevhenii-pyl, line 7: .then(() => fetch(BASE_URL + url))

src/pages/HomePage/components/Categories/Categories.tsx
```diff
@@ -0,0 +1,34 @@
+import { CategoryCard } from '../CategoryCard';
+import './Categories.scss';
+
+export const Categories: React.FC = () => {
+  const phoneModelsAmount = 3;
```
yevhenii-pyl, line 5: Looks like a magic numbers here. Let's imitate env variables, by declaring these limits outside the component scope in uppercase.

src/pages/HomePage/components/PicturesSlider/PicturesSlider.tsx
```diff
@@ -0,0 +1,53 @@
...
+import classNames from 'classnames';
+import { Tabs } from '../../../../types/Tabs';
+import './PicturesSlider.scss';
+
+export const PicturesSlider = () => {
+  const [activeTab, setActiveTab] = useState<Tabs>(1);
+  const iconSrc = './icons/slider-icon.svg';
+  const activeIconSrc = './icons/slider-icon-active.svg';
+
+  const bannerClass = useMemo(
```
yevhenii-pyl, line 11: Why do you want to memoize it?

yevhenii-pyl (changes requested):
Keep on pushing a good job! 
Just a small adjustment with url will do the trick here.

cypress/integration/page.spec.js
```diff
@@ -4,6 +4,6 @@ describe('App', () => {
   });
 
   it('should have h1', () => {
-    cy.get('h1').should('have.text', 'Product Catalog');
+    cy.get('h1').should('have.text', 'Welcome to Nice Gadgets store!');
```
yevhenii-pyl, line 7: 👍

yevhenii-pyl:
Double-check your header, for mobile version it should stretch from 320px to639px screen width: 
(try using `fr` properties for columns, instead of hardcoded `px`)

nav-d
nav-o

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,115 @@
+import { Link } from 'react-router-dom';
+import { useEffect, useState } from 'react';
+import './Header.scss';
+import '../../styles/main.scss';
+
+export const Header: React.FC = () => {
+  const [isMenuVisible, setIsMenuVisible] = useState<boolean>(false);
```
yevhenii-pyl, line 7: Why don't you create a custom hook for this logic? 
Like, `useOpenMenu` or something, it's definitely not required, but may be a good practice, just for you to get that feeling of creating custom hooks for logic encapsulation.

src/context/AppContext.tsx
```diff
@@ -0,0 +1,106 @@
...
+  incrementProductInCart: () => {},
+  decrementProductInCart: () => {},
+  removeFromCart: () => {},
+  clearCart: () => {},
+});
+
+export const AppContext: React.FC<{ children: React.ReactNode }> = ({
+  children,
+}) => {
+  const cartFromLocalStorage = localStorage.getItem('cart');
```
yevhenii-pyl, line 26: As an example of a custom hook you may use here: 

```
export const useLocalStorage = (key, initialValue) => {
  const [storedValue, setStoredValue] = useState(() => {
    try {
      const item = window.localStorage.getItem(key);
      return item ? JSON.parse(item) : initialValue;
    } catch (error) {
      console.log(error);
      return initialValue;
    }
  });
```

As you can see:
1. It's safer since you have `try/catch` to work with errors without app crash,
2. It's easier to provide an `initialValue`,
3. It's just easier to use all around your app :)

src/context/FavoritesContext.tsx
```diff
@@ -0,0 +1,40 @@
...
+});
+
+type Props = {
+  children: React.ReactNode;
+};
+
+export const FavoritesProvider: React.FC<Props> = ({ children }) => {
+  const [favoriteProducts, setFavoriteProducts] = useState<Product[]>([]);
+
+  useEffect(() => {
```
yevhenii-pyl, line 21: This would be much easier if only you had a custom hook to work with local storage here, isn't it?.. :)

src/helpers/colors.ts
```diff
@@ -0,0 +1,25 @@
+export const productColors = {
```
yevhenii-pyl, line 1: Nice idea! 
But even better(and more generic approach) will be to create a `theme.ts` file for all stylings, you may store all necessary css variables there and they always will be by hand.

src/pages/HomePage/components/Categories/Categories.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { Product } from '../../../../types/ProductCard';
+import './Categories.scss';
+
+export const Categories: React.FC = () => {
+  const [phoneModelsAmount, setPhoneModelsAmount] = useState<number>(0);
+  const [tabletModelsAmount, setTabletModelsAmount] = useState<number>(0);
+  const [accessoryModelsAmount, setAccessoryModelsAmount] = useState<number>(0);
+
+  useEffect(() => {
+    getProducts(ProductCategories.Phones).then((phones: Product[]) => {
```
yevhenii-pyl, line 14: Good job! 
Just as a sneak-peak in future - usually, backend should return you data in format: 

```
{
  data: [],
  count: 0
}
```

Frontend should never take care for such things, like counting products on it's own( how can you implement a pagination in this case?).

src/pages/ProductDetailsPage/ProductDetailsPage.tsx
```diff
@@ -0,0 +1,240 @@
...
+    category,
+    name,
+    capacityAvailable,
+    capacity,
+    priceRegular,
+    priceDiscount,
+    colorsAvailable,
+    images,
+    description,
+  } = displayedProduct;
```
yevhenii-pyl, line 96: 💪

src/pages/ProductPage/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,117 @@
+import cn from 'classnames';
+import './Pagination.scss';
+import { useEffect, useState } from 'react';
```
yevhenii-pyl, line 3: Imports should be ordered( you know the order), you are already too precise as a developer, to neglect such 'small' things :)

yevhenii-pyl (changes requested):
Good job, I love going through your code. 

I've highlighted a few things to adjust || that may be done even better, other than that - keep doing the great job! Not much left for this project to be done. 💪

BogdanMaliuta (changes requested):
Hello, here is some things to fix:
image
image
Your slider, it changes when changes image inside it, also on mobile version width should be like viewport width.

image
So big left padding/margin here 

Same here wifth right side, your page should be resposive until its gets tablet breakpoint, 
image

olafchuszno:
Mobile version (320-639px) ready

src/App.tsx
```diff
@@ -1,7 +1,46 @@
...
+    <Routes>
+      <Route path="/" index element={<HomePage />} />
+
+      <Route
+        path="/phones"
+        element={<ProductPage category={ProductCategories.Phones} />}
+      />
+      <Route
+        path="/tablets"
+        element={<ProductPage category={ProductCategories.Tablets} />}
```
BogdanMaliuta, line 28: you can destruct your ProductCategories

src/components/Footer/Footer.scss
```diff
@@ -0,0 +1,73 @@
...
+  &__top-link {
+    grid-column: 1 / 5;
+  }
+}
+
+.footer-nav {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  gap: 16px;
```
BogdanMaliuta, line 29: you can use variables for your gaps, it`s good for calculate in future your container width

src/components/Header/Header.scss
```diff
@@ -0,0 +1,198 @@
...
+
+    display: flex;
+    align-items: center;
+    justify-content: center;
+
+    box-sizing: border-box;
+    height: 48px;
+    width: 48px;
+
+    // margin-right: -17px;
```
BogdanMaliuta, line 37: remove commetns from your code

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+            <nav className="menu-category-nav menu-top__nav">
+              <Link
+                className="menu-category-nav__link"
+                to="/"
+                onClick={hideMenu}
+              >
+                Home
+              </Link>
+              <Link
```
BogdanMaliuta, line 64: add empty lines here

src/helpers/getProducts.ts
```diff
@@ -0,0 +1,17 @@
+import { Product } from '../types/ProductCard';
+import { ProductCategories } from '../types/ProductCategories';
+
+const BASE_URL =
+  location.hostname === 'localhost'
+    ? ''
+    : 'https://olafchuszno.github.io/react_phone-catalog';
```
BogdanMaliuta, line 7: in my opinion beter to not use here ternary with empty option

BogdanMaliuta (changes requested):
Nice phone version, I have some suggestions for your code.
Also if you have time you can add animations to your sliders.

BogdanMaliuta (approved):
Hello, very good solution but I have a few things to fix: 
image
Look at "-" and "+" button , its not in the column.
In favorites if I have 2 products its looks like this, maybe needs to change here justify content?
image
So I think its all
Also you can add loaders if you want, smooth animations in catalogs, etc.
But generally work looks Good

PR #411, opened 2024-05-18
Develop

zvir91:
https://zvir91.github.io/react_phone-catalog/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page!
1. Add a favicon on the page 
image

image
image

3. Cards in the bottom row do not need to be centered on the page
image

4. Also you need to fix font styles everywhere on the page, check it on the design 
image

5. You need to update total for products 
image
image
image

volodymyr-soltys97 (changes requested):
Hi, the comments from the previous review are still not fixed, you need to make a deploy after your changes

TarasHoliuk (changes requested):
Check comments

SanyaBratashchuk (approved):

PR #431, opened 2024-06-06
Develop

AlinaOvod:
https://alinaovod.github.io/react_phone-catalog/


SanyaBratashchuk (changes requested):
1. handle plural it should be `1 item` but  `2 items`
image
2.  apply `cursor: pointer` for buttons and other interactable elements

SanyaBratashchuk (changes requested):
Almost done! 
1. Get rid off horizontal scroll on smaller resolutions

anastasiiavorobiova (approved):
Great work!

PR #432, opened 2024-06-06
Develop

isaeev:
[**Link**](https://isaeev.github.io/react_phone-catalog/)


SanyaBratashchuk (changes requested):
1. check layout with design 
image
2. there are 11 items 
image
3. set 1 page as active on `items on page` change
4. footer should be stick to the bottom 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. Add a favicon on the page 
image

3. Add cursor pointer for all clickable elements 
image

4. You need to update the number of products at the top right next to the icon
image

5. The images shouldn't overlap each other when hovering 
image

6. When going to the page with the phone, we should be transferred to the beginning of the page
image

SanyaBratashchuk (approved):
LGTM

PR #436, opened 2024-06-12
solution

andreyysak:
[DEMO LINK](https://andreyysak.github.io/react_phone-catalog/#/)


anastasiia-tilikina (changes requested):
still not working(

if you have problems with that - feel free to ask your questions in chat)

TarasHoliuk:
1. Add favicon

2. The last image in the slider doesn't exist:

image

3. Check all fonts (weight, color, size) in the card. Design:

image

App:

image

4. This link leads to a 404 page (the page itself looks nice BTW 🙂 ):

image

5. The footer always should be at the bottom of the page:
image

TarasHoliuk (changes requested):
Looks great! Let's improve it a bit to make even better

DarkMistyRoom (changes requested):
Good job! I found a few more points
- would be great to change the button's color when we add the item to the cart
image

- Consider making pagination via `...` like this
image

- hide the pagination when the `All` option is selected
- implement the search 
- white color is chosen, but the phone is black
image

- add a favicon

vadiimvooo (changes requested):
Seems like you forgot to deploy your solution after fix. Please do it and rerequest the review.

VitaliyBondarenko1982 (approved):
Looks good to me

PR #439, opened 2024-06-14
Develop

lmuias:
https://lmuias.github.io/react_phone-catalog/


l4st1m0za (changes requested):
1) Button should be in one lever
image
image

2) Some thing is wrong here
image

3) After I have restarted the page everything dissappeared from the liked and cart

lmuias:
Can you tell me , where you have found this problem with buttons(component and page extension)? Because i checked again and everything is fine even at 320px screen width, maybe i have missed something.

l4st1m0za (changes requested):
Starting from 1200px you have a problem with buttons
image

Also, if you have any questions, please ask in FE chat!

volodymyr-soltys97 (changes requested):
Still not fixed from the previous review, you need to make a deploy after your changes 
image

Check font styles here as on the design 
image

VitaliyBondarenko1982 (approved):
Looks good.

PR #441, opened 2024-06-15
Develop

Maksym-Atanasov:
[DEMO LINK](https://maksym-atanasov.github.io/react_phone-catalog/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better
1. Add a favicon on the page
image

2. You need to fix count for total 
image

3. You need to show the quantity here next to the icons
image

4. When you click on add again, it will be better to remove the product from the cart
image

5. Add transition when this menu appears on the page 
image

6. The content, together with the cards, should stretch to the full available width
image

vadiimvooo (changes requested):
Good job, a few changes required:
1. This images should not overlap each other.
image
2. Add possibility to remove phone from cart by pressing the button
image
3. Something wrong with products pages, please fix it.
image
4. The same for favourites page
image
5. Add spacing between title and buttons
image
6. Add some spacing here as well
image

PR #443, opened 2024-06-19
Develop

io-med:
initial implementation


BudnikOleksii (changes requested):
Looks good 👍 
Here are a few moments that should be fixed or can be improved:
- When a user clicks on "Added to cart" it adds a product to the cart, but the user can't undo it. The only way that I've found to remove the item from the cart is to do it inside a cart.
- When a user clicks on the product page on some color/capacity nothing happens, but actually you should show other product. 
- Optional. I'm not sure is it a feature or not, but I can't swipe sliders on the laptop, they only work on tabs and mobile form.
- On mobile footer looks strange:
image

PR #450, opened 2024-06-28
Develop

yaroslav-na:
[DEMO](<https://yaroslav-na.github.io/react_phone-catalog/>)


SanyaBratashchuk (approved):
1. handle plural it should be `1 item` but `2 items`
image

PR #451, opened 2024-07-02
work

FoReWwEr:
https://FoReWwEr.github.io/react_phone-catalog/


etojeDenys (changes requested):
1. cannot open items from this slider
2. last item in this slider looks cropped
image


3. looks different
image
image

4. add the favicon
5. should be homepage, not phones in this case

https://github.com/mate-academy/react_phone-catalog/assets/70220667/6dabd395-cc5d-47c1-8779-0e8d6c15f63a

SanyaBratashchuk (changes requested):
1. add badge with count of added items 
image
2. handle plural it should be `1 item` but `2 items`
image

volodymyr-soltys97 (changes requested):
Good job!
Let's improve your code 
1. When you increase or decrease the number of products in the cart, this information about count is not updated here
image
image
image

image

3. The structure is now incorrect, the a tag should be inside the li tag, also the height of the links should be 100% of the header height 
image

4. Add transition when the burger menu appears on the page 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. The height of these links should be 100% of the header height. Also add transition for hover effects here
image

2. These comments still not fixed, if you need help feel free in the fe_chat 
image

vadiimvooo (approved):
Well done

PR #452, opened 2024-07-02
react_phone-catalog

IShamkii:
DEMO LINK [https://ishamkii.github.io/react_phone-catalog/]


SanyaBratashchuk (changes requested):
1. handle plural it should be `1 item` but `2 items` image
2., image is not displayed
image
3. remove from card/favorites on second click on button

volodymyr-soltys97 (changes requested):
Hi, if you not implemented the theme switcher, consider will be better of you hide this block 
image

Add the your github profile link to rights links 
image

GUSILLUS (changes requested):
GJ!

1. Consider creating logic to not show button if we don't have any scroll
image

volodymyr-soltys97 (changes requested):
Almost done!
To improve
1. It will be better if when you click add again, you will remove the product from the basket
image

GUSILLUS (changes requested):
GJ! 

1.  Make this button toggleable same to add to cart
image

SanyaBratashchuk (approved):

PR #457, opened 2024-07-07
Task solution

valost:
https://valost.github.io/react_phone-catalog


valost:
Hi! I just saw that the product details page doesn't work for some reason. I'll check what happened and fix it.

valost:
To be honest I am confused why this happened. I was fixing the lint errors and checking if the catalogue was still working. Everything seemed to work fine. Any ideas why this happened?

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon and actual title on the page
image

image

3. When we increase or decrease the product in the cart, the number in the top right next to the cart icon is not updated
image

4. If you click on the product here, you get the following error
image
image

PR #459, opened 2024-07-08
task done

Reaffith:
[DEMO LINK](https://Reaffith.github.io/react_phone-catalog/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. The height of the links should be 100% of the height of the header
image

image

3. When increasing or decreasing the number of products in the basket, this number is not updated in the top right next to the basket icon
image

4. There should be no retreat here
image

5. When clicking on the image, nothing happens, there should also be a transition to the product page
image

6. Fix font styles everywhere on the page, check it on the mockup

volodymyr-soltys97 (changes requested):
GJ!
To improve:
1. This comment still not fixed from the previous review, fix this block as on the mockup
image
image

2. When we open a page with a product, this page should automatically scroll up
image
image

3. These buttons aren't working 
image

4. Add hover effects for all clickable elements on the page
image
image
image

VitaliyBondarenko1982 (approved):
Looks good.
Still not fixed one moment
Screenshot 2024-07-13 at 16 48 41

Suggest to fix it during prepare project for portfolio

PR #461, opened 2024-07-10
Develop

MariiaSuper:
https://MariiaSuper.github.io/react_phone-catalog/


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon and title on the page
image

image
image

3. The height of these links should be 100% of the header height
image

4. When our page is loaded, it is better to show a spinning spinner instead of the loading text
image

5. Add cursor pointer for all clickable elements 
image

6. When we increase or decrease goods, the quantity here is not equal to the price
image

volodymyr-soltys97 (changes requested):
GJ!
To improve:
1. This comment stills not fixed from the previous review 
image

2. When clicking on a product from this list, the page should automatically scroll up when the product changes
image

3. Add here link to your github profile 
image

VitaliyBondarenko1982 (changes requested):
Good job.
A couple moments should be fixed.
- Limit content width by design.
- Exists horizontal scroll on any page for any width.
Screenshot 2024-07-11 at 06 20 53

MariiaSuper:
I have no horizontal scrolling on any page at any width. what browser are you using?

volodymyr-soltys97 (changes requested):
Good job!
To improve
1. Now we have 29 units of the product together, but only 2 are shown under the price, this should be corrected
image

2. The buttons and the block with the price must be aligned everywhere as it is on the layout
image
Layout 
image

3. Add hover effects here 
image
image

lerastarynets (approved):
lgtm 🚀

PR #462, opened 2024-07-12
phone catalog

ivan-baranovskyi:
[DEMO](https://ivan-baranovskyi.github.io/react_phone-catalog/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,13 @@
-// not empty
+@use './shared/global/utils/mixins' as *;
+
+.page {
+  display: flex;
+  flex-direction: column;
+  min-height: 100vh;
+
+  &__container {
+    flex: 1;
+    margin-top: 24px;
+    margin-bottom: 80px;
+  }
+}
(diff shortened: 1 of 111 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. Remove default blue border when input is focused
image

2. When we increase or decrease the products in the basket, we need to update the total amount in the top right next to the basket icon and below the price
image

3. Add the your github profile link for these elements 
image

4. There are no images in slider on the mobile version 
image

5. Remove the scroll when the burger menu is open 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Now we have selected 45 units of the product, but only 20 are displayed under the price and on the top right near the basket icon, this needs to be fixed 
image

2. Open this link in the new tab 
image

3. When we open any page with a product, it should automatically scroll to the top so that the product is immediately visible
image

4.

anastasiiavorobiova (approved):
Great work!

PR #465, opened 2024-07-15
react-phone-catalog 

svyatoslavPy:
https://svyatoslavpy.github.io/react_phone-catalog/


TarasHoliuk:
1. Add the favicon
2. Icons here should display current number of products in cart/favorites:
image

3. There are 8 items. Taking into account that we display the total sum it might be confusing to show the number of unique items.
image

4. Bug with pagination. Select 4 items on the page ->  select the last page -> select 16 items on the page. Result:
image

Solution: After changing number of items on the page set 1st page.
Also, you shouldn't display all pages (31) in pagination. Simple version - display a few closest pages. Improved variant - a few closest + the last page .

5. The logo (in the footer too) should always be a home link:
image

6. It would be nice to have the possibility to remove it from the cart when you click again (similar to favorites). On the product card too.
image

7. It should be link:
image

8. There should be a menu button in the mobile version (it should be visible starting from 639px width):

image

TarasHoliuk (changes requested):
You've done a great job!

But some fixes are required (comments)

Also, fix lint errors:
image

anastasiiavorobiova (changes requested):
Great work! 
Screenshot 2024-07-22 at 16 31 50
The header is scrollable here. Consider fixing it 
Screenshot 2024-07-22 at 16 31 20
Dots shouldn't be clickable

Also, consider fixing lint and test errors

SanyaBratashchuk (approved):

PR #469, opened 2024-07-17
react_phone-catalog

Donovanoff:
[DEMO](https://donovanoff.github.io/catalog-app/)


volodymyr-soltys97 (changes requested):
Hi, good start with header.
Keep going with next sections and page

PR #473, opened 2024-07-22
add task solution

YevhenProtasov:
[DEMO LINK](https://yevhenprotasov.github.io/react_phone-catalog/)


anastasiiavorobiova (changes requested):
Great work! Consider fixing test and lint errors

YevhenProtasov:
I added a rule to .stylelintrc.js so that it doesn't throw an error on using :global
I use global classes to animate some blocks where I couldn't do it with just module classes.
If there is any other solution, I would be grateful if you could provide a link with a possible solution. Thanks

TarasHoliuk:
Wow! Language and theme switch 👍  One of the best Phone catalog I ever seen 🔥 

I cannot reject it 🙂 But I want to add 3 small suggestions (for real too tiny to reject because of them). You can add corresponding changes and push it here but do not re-request the review. Or you can just add these changes later, in the clean repository which you will add to your portfolio (don't worry - there will be a separate task for it).

So, suggestions:
1. Use SVG image for logo for better quality.
2. Make breadcrumbs parts links (i.e link to Phones page in this case):
image

3. Open the GitHub page in the new tab (it is related to all external links):
image

TarasHoliuk (approved):

PR #475, opened 2024-07-24
React phone catalog page

BodyaKutsyk:
[page link](https://bodyakutsyk.github.io/react_phone-catalog/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. You need to fix these errors 
image

2. Add a favicon on the page 
image

3. Remove this b container for all links 
image

4. Add cursor pointer for all clickable elements on the page

5. When we increase or decrease products in the basket, this number is not updated near the basket icon at the top right
image

6. Add transition when the burger menu appears on the page 
image

TarasHoliuk:
1. I selected 4 items on the page and then selected the last (31) page. After selecting 16 items on the page got this:

image

Set page 1 on `Items per page` select change

2. Point 5 from the previous review:

image

3. Light theme - Product name barely visible:
image

4. Optionally. It would be nice to add some padding to these links. Now it's difficult to click them:
image

TarasHoliuk (changes requested):
You did a great job! BTW, skeleton looks nice 👍 
Let's add a few more fixes/improvements

etojeDenys (changes requested):
1. It is better to show some loader in this case instead of a message that the product is not found

https://github.com/user-attachments/assets/8b01469c-4411-4aa7-bada-e29224ad399a


2. if we have more than 5 pages, it's better to show "..." and last page or something like that
image

3. minus looks here as disabled
image

GUSILLUS (changes requested):
GJ! 

1. Consider aspect ratio 
image

2. Consider add disable status or make buttons toggle able 
image

3. Consider don't show bact to top button if you don't have scroll
image

4. Consider fixing UI of cart component
image

volodymyr-soltys97 (approved):
Great work 🔥

PR #477, opened 2024-08-04
Solution to the task

Karanelus:
[DEMO LINK](https://karanelus.github.io/react_phone-catalog)


Karanelus:
Have a problems with font adding from folder =(

BogdanMaliuta (approved):
For now, only fix your header right buttons(cart, liked), and keep working

Karanelus:
Images is not loading even with process.env.PUBLIC_URL
How to fix that?

PR #482, opened 2024-08-15
develop

yuriiyepikhov:
https://yuriiyepikhov.github.io/react_phone-catalog/#/


etojeDenys (changes requested):
image
in this pagination, it is difficult for the user to understand that this is not the last page, because it looks like that


2. add the logic to scroll up in such cases

https://github.com/user-attachments/assets/81a79ab0-1c50-4f97-a641-cbbb0bf96c1c

etojeDenys (approved):
good job

PR #485, opened 2024-08-19
Develop

Nazarin565:
Demo https://nazarin565.github.io/react_phone-catalog/


volodymyr-soltys97 (changes requested):
Good job!
Let's improve your code
1. When we increase or decrease products in the cart, the number in the top right next to the cart icon is not updated 
image
image
image

2. Add the your github profile link for these elements 
image

3. Remove default border for buttons 
image

4. Fix the image position in slider on the tablet version  
image

etojeDenys (approved):
looks awesome

PR #487, opened 2024-08-20
Solution to task

Karanelus:


Karanelus:
Problem with Brand new slider: It's have overflow hidden, but size of page is increased and looks so wide

Zibi95 (changes requested):
Did you deployed after recent changes? For me it looks like that right now 
image

All i can say right now that nav is working and looks ok and banner slider logic is fine. Good job.

Karanelus:
> Did you deployed after recent changes? For me it looks like that right now image
> 
> All i can say right now that nav is working and looks ok and banner slider logic is fine. Good job.

Banner must looks like that, but the rest of the app are not loaded (local host showing all fine)
Send you a video with my problem

https://github.com/user-attachments/assets/e87bd2f5-27c2-4cb8-a8d5-91920969b71d

Karanelus:
` <section className="w-full overflow-hidden">
      <section
        style={{
          width: `${newProducts.length * (276 + 16) - 16}px`,
        }}
        className="grid grid-flow-col gap-8"
      >
        {newProducts.map((phone) => (
          <ProductCardProduct key={phone.id} product={phone} />
        ))}
      </section>
    </section>`
    
    Something wrong here, and i can't handle it rn(

Zibi95 (changes requested):
If I could just have it in the demo i could do more inspection on that. What i would try first is to setup overflow-x: hidden on the body.

Karanelus:
> If I could just have it in the demo i could do more inspection on that. What i would try first is to setup overflow-x: hidden on the body.

Was done earlier, it's only block scrolling to section ('cos it is wide and placed far away from left side of the page).

Karanelus:
Need a help with Home Banner part. It change size of page (not as much as slider earlier, but inflict the navigation)

src/index.css
```diff
@@ -0,0 +1,71 @@
...
+html {
+  overflow-x: hidden;
+  width: 100dvw;
+}
+
+#root {
+  min-height: 100dvh;
+  display: grid;
+  grid-template-rows: auto 1fr auto;
+}
```
natalia-klonowska, line 36: I think problem lies not with banner but with this grid. You can try to fix it or completely delete it and then adjust the rest

natalia-klonowska (changes requested):

src/index.css
Karanelus: Grid was not a problem, but solution =) It fix my layout in good way and stopped change page wigth

Zibi95 (changes requested):
Hey. 
Home view problems
- adjust sizes of banner images for lower resolutions (look on mockup)
- contrast here is bad 
{552DCA85-A2C9-4D7E-A7BC-8ADBB2D595D9}
- add swipe so we can change images of banner on mobile
- back to top button sends me to 404 page
- this title on mobile should be centered 
{E876D56A-A766-487F-8947-1DB386D69951}
- There is no slider for Hot prices section 
{6CB374EA-06EA-4932-92F3-606D08C243E5}
- when page is loading there is layout shift on this section - add some loading spinner and add some height so we won't see the shift
{4FDF5FA1-C3C3-49A6-9FA6-4DC6D738683C}

src/context/AppContext.tsx
```diff
@@ -0,0 +1,223 @@
...
+      switch (sortType) {
+        case "Newest":
+          return b.year - a.year;
+        case "Alphabet":
+          return a.name.localeCompare(b.name);
+        case "Expensive":
+          return b.fullPrice - a.fullPrice;
+        case "Cheaper":
+          return a.fullPrice - b.fullPrice;
+      }
```
natalia-klonowska, line 194: sort is not working cos sortType has additional quotation marks. These extra quotes prevent sortType from matching any of the expected values. if you replace them by e.g sortType.replace(/^['"]|['"]$/g, ''), sorting will work properly ;)

natalia-klonowska (changes requested):

danon321 (changes requested):
image
Can see Hot prices on mobile but cannot on desktop :(

src/components/Home/HomeBanner/HomeBanner.tsx
natalia-klonowska, line 1: images inside HomeBanner don't adjust correctly when screen size changes
image
image

src/components/Phones/Phones.tsx
natalia-klonowska, line 1: phones grid should work like in tablets and accessories pages  
image
image

src/components/Home/HomeHotPrices/HomeHotPrices.tsx
```diff
@@ -0,0 +1,48 @@
+import { useEffect, useState } from "react";
+import { useAppContext } from "../../../context/AppContext";
+import HomeHotPricesButtons from "./HomeHotPricesButtons";
+import Slider from "../../../UI/Slider";
+import useWindowSize from "../../../hooks/useWindowSize.hook";
+
+const HomeHotPrices = () => {
+  const { productsList } = useAppContext();
+  const [curElem, setCurElem] = useState<number>(0);
+  const itemsList = productsList.filter((el) => el.price / el.fullPrice < 0.75);
```
natalia-klonowska, line 10: change filtering for hot prices section so there would be more products than 3 watches ;)

src/components/Cart/Cart.tsx
natalia-klonowska, line 1: cart page content is cut off and you should also add some kind of checkout functionality 
image

natalia-klonowska (changes requested):

danon321 (approved):
Git jest mojonez

PR #489, opened 2024-08-22
Develop

yurovych:
https://yurovych.github.io/react_phone-catalog/


TarasHoliuk:
1. Fix lint errors (try to figure out each error, google it, and possible solutions; if cannot solve the problem disable the rule, or ask for help in the chat):

image

2. Same with ESlint:
image

3. Add favicon

4. Bug with empty displayed pagination:
image

To reproduce:
- select 4 items on the page
- in pagination "scroll" to the last page and select it
- select 16 items on the page
- observe the bug.

5. Use SVG for a logo (in light theme too) for better quality:
image

6. These buttons should work:

image

Change the current product ID based on color and capacity and set it as path (i.e. buttons should be links)

7. it would be nice to remove the product from the cart when click again (same on product details page):

image

TarasHoliuk (changes requested):
Wow! Theme and language change looks great!

But there are things that should be improved/fixed

GUSILLUS (changes requested):
GJ!

1. Fix hover overlapping 
image

2. Consider making 4 by default
image

3. You don't need this button if you don't have scroll, also hide scrollbar
image

4. Consider improve theme toggle UI
image

vadiimvooo (changes requested):
Good job, a few things to improve:
1. Remove pagination on load
image
2.There is a bug when on first load there are no footer
image

yurovych:
Can you try again to check the bug with "no footer on first load", because I think it happend as an accident, cause I try it on different devices and I don't have any problem with this.
All the rest was fixed

GUSILLUS (changes requested):
GJ!

1. Center items and reduce gap
image

2. Arrow button need to change pages not to scroll it
image

3. Don't have scroll until I click on the page
image

etojeDenys (changes requested):
looks good, but I cannot open the link to some product
https://yurovych.github.io/react_phone-catalog/#/phones/apple-iphone-14-256gb-midnight

can you do that?

anastasiiavorobiova (approved):
Awesome!

PR #490, opened 2024-08-25
first-phone-catalog

Klimukalina:
https://klimukalina.github.io/react_phone-catalog/


GUSILLUS (changes requested):
GJ!

1.Consider increase step og slider form 1 to 4 per time
image

2. Make 4 by default
image

3. Try to center name of page
image

4. Why you need this button if you don't have any scroll?
image

GUSILLUS (changes requested):
GJ!

1. Fix logic to show only 4 item by default
image

2. Make this workable
image

3. Don't have scroll but have button
image

volodymyr-soltys97 (changes requested):
GJ!
To improve:
1. Add a favicon on the page 
image

2. Open this additional link in the new tab 
image

etojeDenys (changes requested):
1. better to display some difference between the disabled button and the active one
image
2. don't you want to implement those selectors?
image

GUSILLUS (changes requested):
GJ!


But try to fix your demo link
image

etojeDenys (changes requested):
when I add some product to the cart and reload the page, the button looks like this product is not in the cart
https://github.com/user-attachments/assets/6c9a9b3a-f183-42b3-a454-17ff79a9dfcd


2. better to implement it
image

VitaliyBondarenko1982 (changes requested):
Almost  done.
Needs some fixes on product details page.
- available color and capacity should be clickable and lead to product page with such parameters.
- favorite button not works on product details page

https://github.com/user-attachments/assets/db7c9ae2-7d71-405d-8c0f-9be01c67ef28

- looks like broken logic on click to small product images. they shouldn't change, after click should show selected image on big image, selected image should have have selected styles (see design)

https://github.com/user-attachments/assets/6324b562-fbf7-473b-bd20-2c12c871ac68

- add to cart button also should remove product from cart.

PR #492, opened 2024-08-28
Develop

Miishca:
https://miishca.github.io/react_phone-catalog/


TarasHoliuk (approved):
Great job! Approved ✔️ 

One small suggestion (just fix it before adding to the portfolio, no need to re-request review):

- Prices without discount should be crossed out:
image

PR #494, opened 2024-09-02
Phone catalog

NastyaSid:
[DEMO LINK] (https://nastyasid.github.io/react_phone-catalog/#/)


TarasHoliuk:
1. The bug observed:
- select 4 items/page
- select 31 page
- start search
- observe result:
image

It would be nice to set 1st page when the search input changes (or when the number of pages changes after input change)

2. Remove the item from the cart when clicking this button again (now it adds duplicates to the cart):
image

3. These numbers should always be the same:
image

TarasHoliuk (changes requested):
It works and looks really great!

But a few improvements are required

anastasiiavorobiova (approved):
Great work!

PR #498, opened 2024-09-07
Develop

ivankovbohdan:
[DEMO LINK](https://ivankovbohdan.github.io/react_phone-catalog/).
[Original Dark](https://www.figma.com/file/BUusqCIMAWALqfBahnyIiH/Phone-catalog-(V2)-Original-Dark).


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. Add a favicon on the page 
image

2. When we increase or decrease products in the basket, the number here is not updated
image
image
image

3. Remove scroll when this menu is open 
image

ivankovbohdan:
About 3 - sorry, but I have this function and it seems like it works🤔
photo_2024-09-07_16-51-19

GUSILLUS (changes requested):
GJ!

1. Consider increase ste[ to 4 per time here
image

2. Remove this button if you don't have any scroll
image

3. Improve UI
image

TarasHoliuk:
Have you deployed your solution again after fixes? Because now all previous issues are still reproducible

Also, observed a few more things to improve:

1. Add a favicon (it was in the first review, BTW)

2. These buttons look a bit strange now:

image

Make each 50% width (preferable variant) or at least each 100% width.

3. Set 1st page in pagination on search:

- Select some page with big number:

image

- Type something in search and observe the result:

image

TarasHoliuk (changes requested):
GJ 👍 

Make sure you deployed your solution after code changes

GUSILLUS (changes requested):
GJ!

1. Consider adding background here
image


2. Still have back to top button without any scroll on page
image

3. Improve readability of UI
image
image

4. Same with background
image

5. Show total price for this item
image

ivankovbohdan:
"Still have back to top button without any scroll on page" - I see, I am sorry but I thought that if I just remove it - there will be just an empty space and it will look not really good🤔 So I made the button disabled when there is no scroll and added the appropriate styles to it - and this also applies to the point about readability and this button at the screen, because that's how it was intended - so that it can be seen that the button is not active, it was my idea
image

GUSILLUS (changes requested):
Still have button back to top)

Try to fix this it will improve UI of your project

ivankovbohdan:
> Still have button back to top)
> 
> Try to fix this it will improve UI of your project

Ohh, sorry I just left new comment(not reply) - can you please check my comment a little higher in the general comment stream and tell me please what do you think about that solution with disabled button? Would it be better or should I still remove the button completely?🤔

GUSILLUS (changes requested):
Better to ask your question in the fe_chat

Github pr is not a best place to discussion problem

If you have some problem feel free to ask help in your fe_chat

etojeDenys (approved):
good job

PR #503, opened 2024-09-20
Develop

mixelio:
https://mixelio.github.io/react_phone-catalog/


volodymyr-soltys97 (changes requested):
Hi, the demo link isn't working, you need to make a deploy and check it after on the demo 
image

PR #504, opened 2024-09-20
Develop

MarkiMark3:
DEMO: https://markimark3.github.io/react_phone-catalog/

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 /build
 /node_modules
+/src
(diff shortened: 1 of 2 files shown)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. Add a favicon on the page 
image

image

3. Links should stretch to the full height of the header to make it easy for users to click
image

4. Fix this block, check it on the design 
image

SanyaBratashchuk (approved):

PR #508, opened 2024-09-26
ds

jdifek:
[DEMO](https://jdifek.github.io/react_phone-catalog/)


TarasHoliuk:
1. Fix lint errors:

image

2. It seems like you're trying to hardcode card width (in percent) on the home page here exactly as it is in design:
image

On design, this card width is an example of how it might look on this screen size.  But you can try to adapt it - set 100% and, maybe set some max-width and min-width to handle edge cases and make sure it always looks good.

TarasHoliuk (changes requested):
I see you partially implemented the mobile version. Please, re-request review when the entire task will be done

If you have questions or need help ask in the chat 🙂

etojeDenys (changes requested):
1. images in this section don't look right 
image

2. let's fix this case (display the Midnight color not like the white)
image
just create the object with colors and reuse it
```
const COLORS = {
 midnight: #......,
 ............
}
```

3. add some favicon and update the title
image

4. you forgot to implement the pagination
image


5. display the number of favorite products here
image

6. looks strange on desktop
image

etojeDenys (changes requested):
image
still not fixed


2. footer looks wrong
image

3. add some styles to understand what color is selected
image

4. pagination looks broken on mobile
image

anastasiiavorobiova (changes requested):
Great work!
- Screenshot 2024-10-21 at 16 20 59
- Check styles here
- Screenshot 2024-10-21 at 16 20 35
- Screenshot 2024-10-21 at 16 19 55
- Consider removing the label if there are no products here
- Screenshot 2024-10-21 at 16 20 20
- Check the number here
- Screenshot 2024-10-21 at 16 19 19
- No need to show pagination if `all` is selected
- Screenshot 2024-10-21 at 16 18 54
- Check an alignment here
- Screenshot 2024-10-21 at 16 17 54
- Screenshot 2024-10-21 at 16 18 02
- Check the layout for large desktops
- Screenshot 2024-10-21 at 16 17 12
- Consider adding links to your GH

volodymyr-soltys97 (changes requested):
Good job! 
To improve:
1. The text in buttons shouldn't wrap 
image

2. These buttons must be are on the same vertical line 
image

3. All these elements must be links and add hover effects here, also, open these additional links in the new tab 
image

4. Add transition when this menu appears on the page 
image

5. The slider isn't working on the mobile version, I can't change image 
image

anastasiiavorobiova (changes requested):
Make sure to fix previous comments

etojeDenys (changes requested):
items on page all don't work correctly
image
you should show all cards


2. it's better to display number near these icons on mobile too
image

anastasiiavorobiova (changes requested):
Great work!
- Consider fixing linting errors
- Screenshot 2024-11-01 at 17 27 45
- The product quantity change doesn't work correctly in the cart 
- Screenshot 2024-11-01 at 17 24 46
- Also, the desktop has a horizontal scroll

etojeDenys (approved):
👍

PR #516, opened 2024-10-10
Phone Catalog

Igor-Yaremchuk:
[DEMO LINK]( https://Igor-Yaremchuk.github.io/react_phone-catalog/)

I did this project together with: https://github.com/antonshtef


etojeDenys (changes requested):
looks broken
image


fix it and then re-request the review

TarasHoliuk:
1. The current nav link has the wrong styles (also make links 100% height - same as header height, so it will be easier to click them. Same in the footer):

Figma:
image

App:
image

2. In the cart and in the header you should show number of all items, not only unique items:

image

3. Check font family here and in pagination:

image

4. It makes sense to set 1st page on search change. Because now, if the 1st page (or for example 3rd or even 11th) is selected and I type "11" I can get results:

image

But if I select page 12 I see the page:

image

It's better than showing the same (12) page with no items, but it might confuse the user—it looks like no items match the search query at all.

5. Add favicon and page title.

TarasHoliuk (changes requested):
Great job! Theme and language change 🔥 
But a few things could be improved - see comments

anastasiiavorobiova (changes requested):
Great work! 
The link to the category page doesn't work
Screenshot 2024-10-16 at 16 23 23
Screenshot 2024-10-16 at 16 25 05

etojeDenys (changes requested):
you forgot to remove the capacity 

https://github.com/user-attachments/assets/52a0c1d6-85b0-4fa1-80aa-9ab5f1591c32


2. this slider looks broken
image

anastasiiavorobiova (approved):
Awesome!

PR #526, opened 2024-11-04
Develop

vanvalera:
https://vanvalera.github.io/react_phone-catalog/


SanyaBratashchuk (changes requested):
1. handle plural it should be 1 item but 2 items
image
2. align text 
image

SanyaBratashchuk (approved):

PR #535, opened 2024-11-19
Solution

syavaYki:
[Demo](https://syavayki.github.io/react_phone-catalog/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Add a favicon on the page 
image

image
image

3. Add transition for hover effects everywhere 
image

4. Fix gap between these links and add the your github profile link here 
image

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. You need to open this additional link in the new tab 
image

2. Fix the left indent for the container with cards 
image

3. Don't need to center cards, check it on the design 
image

volodymyr-soltys97 (approved):
Great work! 🔥

PR #541, opened 2024-12-06
Develop

vneholiuk:
https://vneholiuk.github.io/react_phone-catalog/


volodymyr-soltys97 (changes requested):
Good job!
Let's improve your page
1. Add hover effects for all clickable elements 
image
image

2. Add the your github profile link for these elements and open it in the new tab 
image

3. Add transition when the menu appears on the page 
image

4. I can't change image, nothing happens when swiping
image

volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. Fix hover effects for buttons. The button disappears when hovered over.
image
image

2. Need to open these additional links in the new tab 
image

3. Also, you need to fix comments from the previous review with slider and menu

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. Add transition for hover effects everywhere 
image
image

2. Add the your github profile link for this element 
image

3. There should be an option to change the image when swiping 
image

4. There should also be horizontal scrolling here.
image

volodymyr-soltys97 (approved):
Great work! 🔥

PR #542, opened 2024-12-07
Develop

Ehulinian:
DEMO https://ehulinian.github.io/react_phone-catalog/


TarasHoliuk:
There is one bug with pagination:

- select 4 items/page
- select the last page
- select 16 items/page
- the result:

image

Set 1 page after changing items/page

Also, the word must be capitalized in the breadcrumbs

TarasHoliuk (changes requested):
Great job! Everything works smoothly 🔥 One small comment

etojeDenys (changes requested):
akmost done

1. make sure these values are the same

Image

2. make this text a link to the product page

Image

3. page number should not reset after the page is reloaded

Image

brespect (changes requested):
Almost done, check this:
1. There should be the "You may also like" section here, on every product details page:
Image

2. All these buttons shouldn`t scroll me to top, when I press select them:
Image

3. All these links should be real links to the existing resourses:
Image

4. Wrong logic of this count, there should be 7 items:
Image

etojeDenys (changes requested):
1. there should be more than two products

Image

2. it would be better if these links opened in a new tab

Image

etojeDenys (approved):
good job

PR #549, opened 2024-12-21
Develop

k-shestakov:
[DEMO](https://k-shestakov.github.io/react_phone-catalog/)


etojeDenys (changes requested):
total items should be 9
image


but looks good in general

volodymyr-soltys97 (changes requested):
Almost done!
To improve:
1. When we increase or decrease the number of items in the cart, there should be a current number in the upper right corner next to the cart icon.
image

2. Add the your github profile link for these elements 
image

VitaliyBondarenko1982 (approved):
Looks good

PR #555, opened 2025-01-04
Develop

an-marryKyslenko:
[Demo link](https://an-marrykyslenko.github.io/react_phone-catalog/)


volodymyr-soltys97 (changes requested):
Good job!
To improve:
1. Add a correct title on the page 
image

2. The background color should be same everywhere 
image

3. Add hover effects for all clickable elements 
image

etojeDenys (approved):
good job

PR #560, opened 2025-01-07
Develop

tonni004:
[DEMO](https://tonni004.github.io/react_phone-catalog/)


volodymyr-soltys97 (changes requested):
Good job!
Let's improve your code
1. Fix width for buttons in the card, check it on the design 
image

2. You should open these additional links in the new tab 
image

3. These buttons should be on the same vertical line 
image

4. If we don't have any products here, then you don't need to show the number 0.
image

5. Remove scroll when the menu is open 
image

etojeDenys (approved):
looks good in general

PR #563, opened 2025-01-15
solution

hsvirina:
https://hsvirina.github.io/react_phone-catalog/


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add a favicon and topical title on the page 
image

2. Add hover effects here 
image

3. Fix position for these buttons, check it on the design 
image
image

4. Add transition when the menu appears on the page and need to center links 
image

volodymyr-soltys97 (approved):
Good job! 👍

PR #570, opened 2025-01-25
Develop

vitorhvieira:
https://vitorhvieira.github.io/react_phone-catalog/


joaorpereira (changes requested):
The following filters are not working

Screenshot 2025-01-27 at 10 07 59

If I add an product to the favourited section and then try to remove it, the product is not removed
Screenshot 2025-01-27 at 10 08 54

Screenshot 2025-01-27 at 10 08 47

vitorhvieira:
@joaorpereira Here in my browser, the filter selection part worked. And in the favorites section, you can remove it by clicking on the heart. this part of the larger rectangle is for adding or removing from the shopping cart

julianFBarbosa (changes requested):
Although the custom selection inputs are having some strange behaviors (like their options appearing on some element focus within the selection itself, not on the entire selection, and not supporting being opened by pressing tab, the whole thing in one big image looks phenomenal, congrats =)

usability proof (no audio): https://youtu.be/8berwm1ATzk

DanielCamposDevX (approved):
The project looks amazing it is really impressive, just some minor advices that i think doenst affect the project that much:
Remember to add hover effects like scale on products just to give some feedback to the user.
Be carefull with page breakpoints so the user can have a better experience in bigger and smaller screens.
Remember to add on transparent element a little bit of color just to insure that the user can really read the text.

PR #573, opened 2025-01-28
solution

manch0ffline:
[DEMO LINK](https://manch0ffline.github.io/react_phone-catalog/)


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add a favicon and topical title on the page 
image

2. The images loos strange, you need to fix it 
image

3. These elements should be links and add hover effects here 
image

4. Remove scroll when the menu is open 
image

maxim2310 (changes requested):

1. plz change title 
image
2. imgs in slider the same and button on it looks like i can click on it 
image
3.these links should be opened in new tab 
image
4. pager looks like its olderady the end of list 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. This value is incorrect. You need to update the values ​​here and also in the top right near the trash can icon. 
image
image

2. I can't change image on slider on the mobile, need to fix it 
image

volodymyr-soltys97 (changes requested):
Hi, this comment still not fixed from the previous review 
image

TarasHoliuk (approved):
Looks great! Approved ✅

PR #588, opened 2025-02-27
Develop

IrynaZahorodnia:
https://irynazahorodnia.github.io/react_phone-catalog/


VitaliyBondarenko1982 (approved):
Nice job.
A couple minor comments you can improve during prepare project for portfolio.
- small images doesn't match to design. Should be squares and with same sizes for diffent products.
Screenshot 2025-03-30 at 17 24 25
Screenshot 2025-03-30 at 17 25 27
- white line under header.
Screenshot 2025-03-30 at 17 24 38

PR #628, opened 2025-04-04
add task solution

olhakostovetska:
[DEMO LINK](http://localhost:5173/vue-todo-app/)


VitaliyBondarenko1982 (approved):
Good job.
Suggest a couple improvements, which you can make during prepare project for portfolio.
- greate main slider. better to make different images for slides.
- slider with products should work same on each move (now sometimes move on one slide, sometimes on several slides)

https://github.com/user-attachments/assets/2b77c8cf-c6a2-4144-9d78-61fcccba99e3

PR #630, opened 2025-04-08
Develop

Vitalii120296:
https://github.com/Vitalii120296/react_phone-catalog/tree/develop


volodymyr-soltys97 (changes requested):
Hi, this isn't demo link, you need to add the demo link from readme.md file and update it in the PR description 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add the topical title and favicon to the page 
image

2. Add cursor pointer and hover effects for all clickable elements on the page 
image

3. Remove zeros when we don't have any items 
image

4. Add the your github profile link for these elements 
image

5. Need to change the text in button after adding the item 
image

6. The total value should be topical everywhere

maxim2310 (changes requested):

1. check and add cursor pointer to all interactive objects
image
2. links should be opened in new tab 
image
3. add some spaces between and add some selected styles
image
4. looks like color and memory selectors don't work 
image

etojeDenys (approved):
good job

PR #632, opened 2025-04-09
Solution 1.0

sborichevskyi:
[DEMO] (https://sborichevskyi.github.io/react_phone-catalog/)


etojeDenys (changes requested):
1. add some favicon and change the title of the page
image

2. fix those images
image

3. same here
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add hover effects for all clickable elements 
image

image

3. Add the your github profile link for these elements 
image

4. This page looks different as on the design, need to fix it 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. This item looks different as on the design, need to fix it 
image

2. Need to open additional links in the new tab 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. These buttons should be are on the same vertical line 
image

2. The value here should be topical, need to fix 
image
image

3. When clicking on a product, there should be a transition to the product page.
image

anastasiiavorobiova (approved):
Awesome!

PR #646, opened 2025-04-22
Phone catalog

andresleao:
https://andresleao.github.io/react_phone-catalog/


raulriato (approved):
You have a few changes that would be nice, but nothing major, your app works fine, so I'll approve it. Good job!

DanielCamposDevX (approved):

PR #665, opened 2025-05-13
Develop

s1rserg:
[Demo](https://s1rserg.github.io/react_phone-catalog)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
1. The total value should be topical 
image
image

2. Need to disable scroll when the menu is open 
image

volodymyr-soltys97 (approved):
Great! 🔥

PR #672, opened 2025-05-20
Product Catalog

iamgranovskaya:
[DEMO](https://iamgranovskaya.github.io/react_phone-catalog/)


maxim2310 (changes requested):

1. loop is not working while clicking 
image
2. if i click on color and then back i expect to come back to previous page but i just revert previous color 
image
3. efter click on memory page is crushed 
image
4.align these blocks 
image
and this minus 
image
5. add transition 
image
6. and add hover effect here 
image

VitaliyBondarenko1982 (changes requested):
Good job.
Consider some improvements.
- add active styles for active color and active capacity (check design)
Screenshot 2025-05-24 at 18 43 15
- after change perPage page should be settled to 1.

https://github.com/user-attachments/assets/3927adee-3682-4547-adf6-193fe3c55ba8

etojeDenys (changes requested):
1. after increasing or decreasing the number on the left, you need to change the number of items on the right
image

2. after pressing again, it should delete
image

iamgranovskaya:
@etojeDenys regarding delete item when click on `Added to cart` button we do nothing according to Task Readme
```
Add to cart button in the ProductCard should add a product to the Cart.
If the product is already in the Cart the button should say Added to cart and do nothing.
```

maxim2310 (changes requested):
Almost there) Just two more things needs to be fixed

1.blocks are jumpling while increasing count 
https://www.loom.com/share/bab3d5f9e1e74c318f02a58a6ea477a5
2. button back should return to previous page, but in your case it returns me to phone page

PR #676, opened 2025-05-24
Develop

hardaira:
https://hardaira.github.io/react_phone-catalog/


etojeDenys (changes requested):
1. make sure that it looks the same with figma mockup
image

2. after resetting the page, it should stay the same
image
image

3. add hover to photos and cards

hardaira:
If you write that my version differes from the version in a mockup, could you specify what is wrong, because I get lost in guesses.

PR #678, opened 2025-05-27
Phone catalog

Serhii-Khobotov:
https://Serhii-Khobotov.github.io/react_phone-catalog/

### DIFF
--- a/src/modules/AccessoriesPage/AccessoriesPage.module.scss
+++ b/src/modules/AccessoriesPage/AccessoriesPage.module.scss
@@ -3,16 +3,17 @@
 .accessoriesPage {
   &__navigation {
     display: flex;
+      align-items: center;
     column-gap: 8px;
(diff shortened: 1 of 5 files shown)


etojeDenys (changes requested):
1. after opening the demo link, this problem occurred
image
2. phones list is empty
image

3. after reloading the page, we encountered an error
image
use HashRouter here

anastasiiavorobiova (changes requested):
Great work!
- Screenshot 2025-05-28 at 19 00 01
- Check the layout here
- Screenshot 2025-05-28 at 18 59 13
- The link in breadcrumbs should be working
- Screenshot 2025-05-28 at 18 58 56
- Check styles here

etojeDenys (changes requested):
1. after reloading, you should save the same page
image

2. don't forget this block
image

PR #685, opened 2025-05-30
Develop

VadimKudrenko:
[DEMO LINK](https://vadimkudrenko.github.io/react_phone-catalog/)

I choose the original design, but changed the logo and naming because I was developing the project for a clean portfolio repository and wanted to add a bit of identity. And I really want to try to do some logo with AI.

I also slightly changed the logic of displaying the catalog. In my implementation, there is one catalog page (not three separate ones for each product category as in the task) you only need to send one request to the server, and then the products are filtered by search parameters even when the product category changes. Plus, the URL search parameters are not reset when switching to another category.
This approach makes it easier to expand the application for other functionality in the future.

Overall, my main goal was to create a project with a clear architecture that would be convenient for another developer to work with and expand in the future. I spent a lot of time on reusability of components / styles, as well as convenience with the use of various functionality, such as using a locale storage, receiving products with different filters, etc.

P.S 
After you click on checkout button in your cart, you will find a small reference to one iconic scene from Mafia II.
I hope you also had a chance to play this game in your childhood ;)

### DIFF
--- a/src/components/BannerCarouselSection/components/BannerCarousel/BannerCarousel.tsx
+++ b/src/components/BannerCarouselSection/components/BannerCarousel/BannerCarousel.tsx
@@ -1,12 +1,13 @@
 import React from 'react';
 import { Swiper, SwiperSlide } from 'swiper/react';
 import { Navigation, Pagination, Autoplay } from 'swiper/modules';
+import { NavLink } from 'react-router-dom';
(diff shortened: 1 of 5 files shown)


etojeDenys (changes requested):
great job

to improve:

1. Here, you should show the number of added and liked products.
image

2. here you should remove the card after removing like
image

PR #692, opened 2025-06-08
Develop

webEsperer:
https://webesperer.github.io/react_phone-catalog/


Zibi95 (changes requested):
Dark mode looks of 
image
image
image

Missing favicon and title 
image


When i click that link in breadcrumbs of the page that I'm currently in it redirects me to not found page. This link should be inactive
image


On width around 1000px in list of products there is empty space on the right maybe center it on the page
image

PR #695, opened 2025-06-12
React phone_catalog

MateuszCieplak:
[Demo link](https://MateuszCieplak.github.io/react_phone-catalog/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Each component is placed in its own folder with:
 - Dynamic product list with:
   - Loader / error states / empty messages
   - Sorting (`Newest`, `Alphabetically`, `Cheapest`)
-  - Pagination (`4 / 8 / 16 / All` items per page)
+  - Pagination (`4 / 8 / 16 / 32` items per page)
 
 ### 📱 Product Details Page (`/product/:productId`)
 


Zibi95 (changes requested):
Hey, great job! The current implementation looks good and functions perfectly. I was even going to suggest adding swipe functionality for the sliders, but I see you've already implemented that. Really well done!

Here are some things that you can implement:

Add a light mode: This would provide users with more display options.
Implement skeleton loading: For example, on the product page, this would improve the perceived loading speed and user experience.

There are small details. That you can improve 

Some buttons have default cursor when hovering over them 
image
image
image

When i click the link that redirects to the page I'm currently in breadcrumbs then it redirects me to the not found page 
image

Not found page with that black bg is bad
image

Zibi95 (changes requested):
Nice. That switch for light mode button could be better. Maybe ask AI to generate something interesting. 
- On my bigger screen there is white are below 
image


What else. Do you feel like something could we written better? Or reused in components? Look for such improvements. For example do yourself code review. And see what can be improved further. Here are two points that i found as I was scrolling thrugh code base. 
- use import aliases https://shivi.hashnode.dev/simplifying-module-imports-in-vite-react-typescript-with-aliases
- I saw at least one className that was using template string for soem conditionals. Make use of classNames for clearer approach.

natalia-klonowska (changes requested):
niektóre zdjęcia produktów są o wiele mniejsze ponieważ ich tło jest o wiele większe:
https://mateuszcieplak.github.io/react_phone-catalog/img/phones/apple-iphone-14-pro/spaceblack/00.webp
https://mateuszcieplak.github.io/react_phone-catalog/img/phones/apple-iphone-14/purple/00.webp
image

postaraj się aby ta różnica rozmiarów była mniej zauważalna

natalia-klonowska (approved):

PR #699, opened 2025-06-17
need to find a problem

DariaFesiun:


etojeDenys (changes requested):
provide a demo link, and then request a review again

DariaFesiun:
[DEMO](https://dariafesiun.github.io/react_phone-catalog/)

etojeDenys (changes requested):
good job

to improve:
1. the card should be a link that opens the product page when clicked
image

2. There should not be an error after refreshing the "phones" page
image

3. fix the footer
image

4. the buttons should be on the same level
image

etojeDenys (approved):
Great job!

PR #700, opened 2025-06-19
Add phone-catalog solution 

JaneShavrukova:
[DEMO](https://janeshavrukova.github.io/react_phone-catalog/)


volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
1. Add a favicon to the page 
image

2. Add hover effects for all clickable elements and your github profile link to contacts link 
image

3. Need to disable scroll when the menu is open 
image

PR #707, opened 2025-06-30
react_phone-catalog

Yaroslav65:
DEMO LINK - https://yaroslav65.github.io/react_phone-catalog


etojeDenys (changes requested):
good job

1. after pressing "Add to Cart," the product was added, but the button did not change
image


2. add a hover effect
image

3. the image is too big
image

4. add a favicon and a title
image

PR #713, opened 2025-07-08
Develop

ogchy:
https://ogchy.github.io/react_phone-catalog/#/


etojeDenys (changes requested):
1. these links should open in a new tab
image

2. button doesn't work
image

3.You should display which photo and color are currently selected
image

4. look style at layout
image

image

5. fix position 
image

6. these numbers should be the same
image


7. make sure this page has the same layout
your:
image

figma:
image

PR #715, opened 2025-07-09
solution

Vlad-2608:
https://vlad-2608.github.io/react_phone-catalog/


etojeDenys (changes requested):
make sure you provide a demo link and then request a review again

PR #718, opened 2025-07-11
Finish the task

KateLis84:


src/index.tsx
```diff
@@ -1,4 +1,9 @@
 import { createRoot } from 'react-dom/client';
 import { App } from './App';
+import { BrowserRouter } from 'react-router-dom';
```
vitalii-bondarenko-ua, line 3:
```suggestion
import { HashRouter } from 'react-router-dom';
```
use HashRouter instead BrawserRouter - it will help to solve issues with routing on gh pages.

etojeDenys (changes requested):
Almost done

1. make sure your follow buttons match this mockup
your:
image

figma mockup:
image

2. it looks different with the layout
your:
image

layout:
image

PR #720, opened 2025-07-13
add task solution

Dvoris923:
[DEMO]https://dvoris923.github.io/react_phone-catalog/
я додав щоб вирішити білий екан і тд там ще немає що перевіряти)


etojeDenys (changes requested):
good job

1. after clicking on this link we see "Page Not Found"
image

image

2. these links should open in a new tab
image

3. logo links should lead to the main page
image

etojeDenys (changes requested):
almost done

1. link in the header leads to an error page and  link in the footer does not work
image

2. center page content
image

PR #721, opened 2025-07-13
solution

Vlad-2608:
[DEMO LINK](https://Vlad-2608.github.com/phone-catalog)


etojeDenys (changes requested):
1. on the phone and tablet versions,the footer and header do not take up the full width
image
image

2. add favicon
image

etojeDenys (changes requested):
the header is good but the footer still looks wrong
your:
image

figma example for the phone version:
image

etojeDenys (approved):
Great job!

PR #733, opened 2025-07-26
Initial solution

Mitlosh:
https://mitlosh.github.io/react_phone-catalog/#/


etojeDenys (changes requested):
1. Slightly increase the images when hovering them
image

2. these links should work
image

3. fix it
image

4. There should be a "You May Also Like" section
image

figma:
image

5. buttons should be on the same level
image

6. Increase the number of items when the user presses the plus buttons
image

7. Also increase the price here
image

8. too much space
image

9. add favicon and title

PR #741, opened 2025-08-02
phone_catalog

ivsasha:
https://ivsasha.github.io/react_phone-catalog/#/


etojeDenys (changes requested):
1. these buttons should be on the same level
image

2. the gap here is too big
image

etojeDenys (approved):
Good job

PR #752, opened 2025-08-15
Develop

Taipan-4ik:
https://taipan-4ik.github.io/react_phone-catalog/#/


etojeDenys (changes requested):
1.  footer should not stretch up
image
image
image


2. these parameters should have selected class
image

3. after reloading, you should keep these values
image
image

4. these numbers should be the same 
image

5. add favicon and title

etojeDenys (changes requested):
almost done

1. fix the problem with the menu icon and images on the phone version
image

2. add favicon and title
image

PR #754, opened 2025-08-17
solution

igols:
https://igols.github.io/react_phone-catalog/

### DIFF
--- a/src/layout/Header/Header.module.scss
+++ b/src/layout/Header/Header.module.scss
@@ -247,3 +247,8 @@
   clip: rect(0 0 0 0);
   overflow: hidden;
 }
+
+// stylelint-disable-next-line selector-pseudo-class-no-unknown
+:global(body.no-scroll) {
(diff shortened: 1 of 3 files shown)


etojeDenys (changes requested):
page looks broken
image

maxim2310 (changes requested):

1. align it 
image

2. add some padding here 
image

3. add some space on top 
image

4. here some troubles with contrast on dark theme
image

5. this button doesn't work 
image

6. cards should have same size
image

7. here should be additional selects and pagination at the bottom 
image

8. back should navigate as to previous page but it doesn't if i click on some 
image

etojeDenys (changes requested):
1. use a better style for the selected page
image

2. it would be better if you showed what "item on page" is selected
image

3. add selected class to image
image

4. fix style here
image

5. these values should be the same
image

6. remove this gap
image

7. add favicon and title
image

PR #757, opened 2025-08-19
project created

Uusuff:


etojeDenys (changes requested):
1. after adding to the cart change the button style
image

2. add hover to images
image

3. open these links in new tab
image

4. increase the price here as well
image

5. these values should be the same
image

etojeDenys (changes requested):
almost done

1. place these buttons on the same line
image

2. this should link to the phone page
image

Uusuff:
I forgot to deploy the project after adding changes

etojeDenys (approved):
great job

PR #760, opened 2025-08-21
Dev

Sofiia13:
- [DEMO] (https://sofiia13.github.io/react_phone-catalog/#/)

- [Original Dark](https://www.figma.com/file/BUusqCIMAWALqfBahnyIiH/Phone-catalog-(V2)-Original-Dark)


h-amster (changes requested):
1. Gaps between category images are too big
image
2. Too many items in products slider, last item looks crooped
image
3. Use styles from design for "Added to Cart" button. Also, after click on this button good should be removed from cart
image
4. Add border on card hover
image
5. Price and specs blocks should be aligned. Also if screen specs string is too long make it shorter and add ",,," in the end of the line
image

PR #763, opened 2025-08-24
fix

tavisua:
https://tavisua.github.io/react_phone-catalog/


etojeDenys (changes requested):
1. these links should open in new tab
image

2. logo links should lead to the top of the page

image

4. add favicon and title
image

PR #764, opened 2025-08-25
Develop

Mihakurochkin:
[DEMO] https://react-phone-catalog-mykhailo-kurochkins-projects.vercel.app/


etojeDenys (changes requested):
good job, and great idea with the settings button!

1. add a hover effect to these links
image

2. these values should be the same
image

3. open these links in a new tab
image

PR #766, opened 2025-08-26
Develop

DeJisk:
[DEMO LINK(https://dejisk.github.io/react_phone-catalog/#/)


etojeDenys (changes requested):
great job

1. after refreshing, we see only one tab, but five is selected
image

2. these values should be the same
image

3. add favicon and title 
image

PR #773, opened 2025-09-04
Add solution

10hokageee:
[DEMO LINK](https://10hokageee.github.io/react_phone-catalog/#/)


etojeDenys (changes requested):
1. the logo in the footer leads to an error page. It would be better if it lead to the top of the homepage
image

2. show selected color
image

3. gap is too big
image

4. make sure the buttons are placed on the same line
image

PR #774, opened 2025-09-05
Beta

DSemeniukV:


etojeDenys (changes requested):
1. footer should not stretch
image

2. add more space to page content
image

3. header and footer should span the full screen
image


4. fix styles here
image
image

5. after opening the device page, the entire page breaks

FaiHamid (changes requested):
Greate job!
Let`s just fix few details:

1. The maximum content width according to figma should be 1136px.
image
image

2. Button 'checkout' looks strange.
image

3.An empty cart should look similar to an empty favorites page .
image
image

PR #776, opened 2025-09-07
Develop

vilich-mykhailo:
[DEMO LINK](https://vilich-mykhailo.github.io/react_phone-catalog/)


vitalii-bondarenko-ua (changes requested):
- Now app looks broken. Use HashRouter instead BrowserRouter
Screenshot 2025-09-07 at 16 58 16

- For fix images path should start with `./`
Screenshot 2025-09-07 at 16 53 35

vitalii-bondarenko-ua (changes requested):
Good job.
Needs improvements
- header better to make on full width of page.
- Save pagination params in the URL ?page=2&perPage=8 (page=1 and perPage=all are the default values and should not be added to the URL.
Same for sort and filter params.
After reload page pagination, sort and filter params shouldn't be lost.

Screenshot 2025-10-25 at 18 37 13

https://github.com/user-attachments/assets/edaa0dd1-8693-4fca-b04a-858a0cf1d538

etojeDenys (changes requested):
problem with the search parameters still not fixed. make sure you complete this step
image

PR #777, opened 2025-09-08
Develop

Oksigenka:
DEMOLINK-[https://oksigenka.github.io/react_phone-catalog/#/]

### DIFF
--- a/src/pages/CartPage/CartPage.module.scss
+++ b/src/pages/CartPage/CartPage.module.scss
@@ -2,6 +2,8 @@
   @include content-padding-in-line;
 
   &__backbutton {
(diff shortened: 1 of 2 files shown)


FaiHamid (changes requested):
Greate job🎉
Everything looks almost as it should, let's just change a few small details:

1.Cursor on 'Back' should look like cursor pointer. Now it looks like cursor for text.
image

2. Changing the capacity should also change the URL.
image

3. When I click on the ‘Checkout’ button, nothing happens. There should be some kind of message and the basket should be cleared.

4.The value of items in cart is not updated immediately.
image
image

PR #780, opened 2025-09-09
Develop

dminikulin:
[Deployed project](https://dminikulin.github.io/react_phone-catalog/)


dminikulin:
In API, iPhone XR White has 5 image addresses, but only 4 images of this model are given inside images directory.

PR #785, opened 2025-09-12
add code

ElisabethPO:
https://elisabethpo.github.io/react_phone-catalog/

### DIFF
--- a/src/modules/Catalog/Catalog.module.scss
+++ b/src/modules/Catalog/Catalog.module.scss
@@ -11,12 +11,21 @@ html, body, h2 {
   box-sizing: border-box;
 }
 
+@media (max-width: 480px) {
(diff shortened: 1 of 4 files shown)


etojeDenys (changes requested):
1. text looks different in the layout
image

2. add some space here
image

3. selected image and color should be displayed by default, not only after clicking
image

4. changing color and capacity work stange

5. here should be buttons
image

example:
image

vitalii-bondarenko-ua (changes requested):
Good job. Almost done.

- on some screen width exists horizontal scroll

https://github.com/user-attachments/assets/6a7318da-a72a-41c5-9381-a41372013415

- broken images on favorite page
Screenshot 2025-10-21 at 16 09 44

PR #787, opened 2025-09-13
Develop

mykytalandar:
https://mykytalandar.github.io/react_phone-catalog/#/

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -20,6 +20,18 @@
   min-height: 100vh;
 }
 
+.h1 {
+  position: absolute;
+  width: 1px;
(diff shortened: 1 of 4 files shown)


etojeDenys (changes requested):
1. fix images problem
image

2. these links should open in a new tab
image

3. after changing the number of items on the page, you should set the page number to 1

4. these numbers should be the same
image

5. increase the price here as well
image

6. footer should not stretch, and the footer should be outside the page content padding
image

etojeDenys (approved):
good job. however, it would be better if you placed the button on the same line
image

PR #790, opened 2025-09-17
Develop

kifont:
https://kifont.github.io/react_phone-catalog/ DEMO


etojeDenys (changes requested):
page looks broken. fix it, and then request the review again
image

PR #793, opened 2025-09-26
Develop

Olena7202:
https://Olena7202.github.io/react_phone-catalog


etojeDenys (changes requested):
good job
1. after changing the phone's color or capacity, the phone is still "added," but it's a different phone than the one I added
image

2. should open in a new tab
image

image

4. center content on the phone version
image

PR #795, opened 2025-09-26
header and carusel

OleksandrY0:
https://oleksandry0.github.io/react_phone-catalog/


etojeDenys (changes requested):
page looks unfinished. keep going and request the review when you finish
image

PR #798, opened 2025-10-01
add solution

michelletani:
[DEMO LINK](https://github.com/michelletani/react_phone-catalog)


pedro-ruas (changes requested):
Since you haven't deployed the app, I cloned your repository, tested locally, and found some issues:
- Most images do not load, I believe the URL/data is missing
- There are several components without styling, i.e.: buttons use the default browser button, which makes the page look a little bad
- Favorite's page isn't working. I tried adding a product to favorites, and the counter says the products is there, but the page stays empty
- Product page navigation does not work. Changing pages keeps the same items and makes the page "jump"
- Products page also has bad styling, this should not be a single column of products

PR #801, opened 2025-10-04
add solution

anastasia22422:
https://anastasia22422.github.io/react_phone-catalog/


etojeDenys (changes requested):
great job
however, there are some problems with the images on the prone version

close menu button
image
and loading
image

PR #866, opened 2026-01-16
Develop

adonch:
[DEMO](https://adonch.github.io/react_phone-catalog/)


maxim2310 (changes requested):


1. all animation with transition works choppy, check why its not work on mouse leave event
2. i think you dont need to show loading here when changing colors and memory 

Image

3. back button has to lead to previous page 

Image

Anton-Kuchmasov (approved):
Amazing job you did!

PR #875, opened 2026-01-28
1st variant

Beliyandr:
https://beliyandr.github.io/react_phone-catalog/

I know that I need to add skeleton and another things (


Denys-Kravchuk9988 (changes requested):
Good job! 

A few things to improve:
1. Please use `.svg` format for icons to have better quality here. And I would recommend to display `header` on whole page size
Image
Image

2. Please remove default styles for button (waste border)
Image
Image
Image

3. Text of icons should be white and border dark
Image
Image

4. Prices should be in a row
Image
Image

5. Please, pay attention for margins
Image
Image

6. Cards should be in a row
Image
Image

7. I would recommend to decrease margin according the design
Image
Image

8. I would recommend to change `>` symbol for icon from design
Image
Image

9. Here silver borders are cuf off. I would recommend to use silver border as dark border is used and when it's active to update the silver color to dark
Image
Image

10. Margin between text and button is bigger than on design
Image
Image

11. One image is absent
Image

12. Icon `$` is absent and I choose 7 items of phone and the total count is still 1. Also the border around checkout block is absent
Image
Image

13. It seems that the counter doesn't calculate price correctly in case:
1. I was on cart page and pressed 7 times on plus
2. Moved back on main page -> choose a new iphone and added it to cart
3. On this page I see 1 Iphone 7 but price is higher than expected
Image

Beliyandr:
https://beliyandr.github.io/react_phone-catalog/

I  fixed your recommendetions

etojeDenys (changes requested):
good job

Image

2. it would be good if the product could be removed from the cart after pressing the button again
Image

3. price placment looks broken

Image

4. you should also include the page number in the search parameters
Image

5. the image is too big at the moment, so scrolling appears

Image

6. These numbers should be identical.


Image

7. add a favicon and a title

Beliyandr:
https://beliyandr.github.io/react_phone-catalog/

I fixed your new recommendetions

Anton-Kuchmasov (changes requested):
Good job!


Would you mind to fix this UI issue?

Image

Anton-Kuchmasov (approved):
Well done!


Don't forget to add link to your GitHub profile in Footer (do it before final review)

PR #903, opened 2026-02-24
https://serhiydmytruk.github.io/react_phone-catalog/

SerhiyDmytruk:


FaiHamid (changes requested):
Hi,
Run `npm run deploy ` and add DEMO_LINK to pull request`s description.

SerhiyDmytruk:
https://serhiydmytruk.github.io/react_phone-catalog/

etojeDenys (changes requested):
good job

Image

2. slightly increase size of images when hovering them

Image

3. these links should not lead to an error page

Image

4. logo images look too small. also, it would be good if the page scrolled up after pressing them
Image

5. after pressing the add button, redirect the user to the product page instead of adding the product to the cart

Image

6. after changing the 'item on page' setting, the page number should be set to 1

Image

Image

7. product does not change when different colours are selected
Image

8. fix image

Image

9. make sure these numbers are the same

Image

10. it would be good to remove the product from the page after pressing the 'Add' button again
Image

Anton-Kuchmasov (changes requested):
Almost done!

To improve:
1. Would you mind to make all cards height fixed? Now it looks sligthly weird in this block:

Image

2. Click on one of those cards (see screenshot above) should scroll page up
3. Height of heart & `Add to cart` buttons should be equal too:

Image

4. When I click on `Add to Cart` button here, App opens full info page, but it expected adding to Cart only:

Image

5. Check out 
https://serhiydmytruk.github.io/react_phone-catalog/#/phones/apple-ipad-pro-11-2021-1tb-spacegray

This memory option (as well as 1 or 2 TB) is available but after choosing it App redirected me to `Not Found` page

SerhiyDmytruk:
Other bugs I fixed!

**Note about "Add to cart" behaviour ** -- prev review: 4. When I click on Add to Cart button here, App opens full info page, but it expected adding to Cart only:

The behavior of AddToCart in Slider or on the category page will not be described in detail.

During the previous review, it was indicated that it is worth buying on the Product page - link. So that's what I did.

https://github.com/mate-academy/react_phone-catalog/pull/903#pullrequestreview-3850026186 

_5. after pressing the add button, redirect the user to the product page instead of adding the product to the cart_

image

etojeDenys (changes requested):
almost done

1. but I feel that you have misunderstood that point. that button should add the product to the cart, but in your implementation it redirects the user instead
Image

2. make sure that these links do not lead to an error page
Image

Image

SerhiyDmytruk:
Thanks for remarks!
1 - btn 'addToCart' now work as toggle.
2 - refactored links in the footer and at all

pls check.

thansk for explanation of issues

FaiHamid (changes requested):
Good progress!
Let`s just fix few details:

1.Pagination not adaptive
Image

2.After changing the sort field or the number of items per page, it should navigate to the first page
Image

SerhiyDmytruk:
Good remarks!

FaiHamid (approved):
Well done 👏

PR #895, opened 2026-02-19
Solution

sashkatmshchk-pixel:
Demo:
https://sashkatmshchk-pixel.github.io/react_phone-catalog/

Hi! Final version ready for review.

All mentor fixes applied:
- newest sorting fixed
- correct back navigation
- dynamic document title update
- UI fixes and cleanup
- project builds and works correctly

Ready for final review

PR #907, opened 2026-02-26
Solution

sashkatmshchk-pixel:
Previous PR #895 was auto-closed after force-push. Branch restored to c9361a7; opening replacement PR.

Demo:
https://sashkatmshchk-pixel.github.io/react_phone-catalog/


2pasha (changes requested):
good job! 👏

here are some improvements: 

1. it's better to implement burger menu on mobile 

Image

2. this section look broken on mobile 

Image

3. looks little bit weird on mobile 

Image

4. it's better to fix some height 

Image
Image

5. implement section "You may also like", which will suggest user some other products

Image

etojeDenys (changes requested):
almost done

1. after selecting a product, the page should scroll up

Image

2. there should be an option to scroll right to see more than four options

Image

3. here too much space

Image

4. it should be a link to the product page

Image

2pasha (changes requested):
good job! 👏

here are some improvements: 

1. after user choosing some of these products, page should be scrolled up

Image
Image

2. add more than 4 products here

Image

3. capacity should also be selected by default

Image

4. from cart and favorites user have ability to visit ProductDetailPage

Image

etojeDenys (approved):
lgtm

PR #911, opened 2026-03-01
Develop

Vitaligna:
[DEMO LINK](https://Vitaligna.github.io/react_phone-catalog/)


Denys-Kravchuk9988 (changes requested):
Excellent work!

A few things to improve:
1. Minor fix here. When I try to search by text `series 5` I can't write space between series and 5
Image

2. Products are supposed to display on the left side according to design
Image
Image

3. Colors here are equal (first and last)
Image
Image

4. On tablet the left border goes outside the navigation buttons container
Image


One recommendation

etojeDenys (changes requested):
1. there is still a problem with the colours

Image

2. disable page scrolling when the menu is open

Image

3. it would be great if you could add a hover effect to the product cards

Image

4. add the option to delete a product from the cart by pressing the button again
Image

brespect (changes requested):
Good progress, let`s check this:
1. I add items to the cart and when I click on it there is no search option, so why does the magnifying glass icon disappear when I have 20+ items? 
It is worth adding such an opportunity there as well:
Image

2. When I select something in this section, the page should scroll smooth to the top:
Image

3. Add the. underline hover effect to this links, and each of them should redirect to existing resources:
Image

Anton-Kuchmasov (changes requested):
Amazing job you did!

One tiny fix and your PR will be approved:

1. Check out styles in the `Footer` navbar:

Image

etojeDenys (approved):
great job

PR #915, opened 2026-03-03
React Phone Catalog 

ProKesha:
Completed React Phone Catalog app

DEMO 
https://prokesha.github.io/react_phone-catalog/


etojeDenys (changes requested):
good job

Image

2. photos should not be cropped
Image

3. make sure that this section looks the same as the layout

Image

layout:

Image


4. place the buttons on the same lines
Image

5. add a favicon and a title

brespect (changes requested):
Good progress, almost done, check this:
1. After select item in this section the page should smoothly scrolls to the top:
Image.

2. Use logo as favicon.
3. All of these links in the footer should be real links to relevant existing resources:

Image

Denys-Kravchuk9988 (changes requested):
Good job!

I would recommend to align the images' height to fit height of side buttons and remove the space below these buttons

Image

2pasha (approved):
lgtm 🔥

PR #916, opened 2026-03-03
Develop

iirk1:
https://iirk1.github.io/react_phone-catalog/


brespect (changes requested):
Good progress, let`s check next:
1. Add some padding to give the user the ability to select this sections, not only by clicking directy on the icons:
Image

2. Add favicon and change title in the browser tab.
3. When I select something in pagination it jumps me to the top of page:
Image

4. All this links should lead to correct resourses:
Image

5. Lets add correct icon for this button, not just "^" symbol:
Image

6. When I select something here it should scroll me smooth to the top:
Image

Also, don`t forget to pass all tests and linter staged, before requesting the mentor's review 😉

iirk1:
3. When I select something in pagination it jumps me to the top of page:
I think it`s a good approach because when you click on another page, you should see the top of the new page with new products.

About the linter and test, Mentor said it`s not necessary to pass the test because it`s only a single test, and I wrote the hidden title,  but I don`t understand why it doesn`t pass. Also, for the linter, I don`t know how to fix the problem with the selector :global.

etojeDenys (changes requested):
almost done

1. it would be better if the whole square were a link

Image

2. add the option to remove the product by pressing the button again

Image

3. after changing the 'item on page', the page number should be set to 1

Image

4. make sure you don't forget to increase these numbers as well

Image

FaiHamid (approved):
Well done!

PR #923, opened 2026-03-07
First Request

Sany8k:
[DEMOLINK]https://sany8k.github.io/react_phone-catalog/#/


etojeDenys (changes requested):
great job, just a few fixes to make

1. it would be better to open this link in a new tab
Image

2. after selecting that capacity on that phone, an error page appears
Image

Image

3. add some arrows to these buttons
Image

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. It's better to add dropdown caret for these selects
Image

2. I would recommend to add `cursor: pointer;` when user hovers this section
Image

Denys-Kravchuk9988 (changes requested):
Almost done!

One tiny bug to solve: 
when I close the select's list the caret is in wrong position
Image

etojeDenys (approved):
lgtm

PR #931, opened 2026-03-10
Develop

KazymyrYan:
**[Live Demo](https://kazymyrYan.github.io/react_phone-catalog/)**


brespect (changes requested):
Good progress, lets check next:
1. Add the space between size selectors and error message.
Image

2. The hover effect should not overflow the another image:
Image

3. All links should be proper reference to existing resourses:
Image

4. The hover effect of this card, merge with the main backgorund, fix this:
Image

5. Thiss button is too small, check the mockup:
Image

6. If you select something in this section the page should smoothly scrolls to top, remove the loader:
Image

7. Add favicon.

8. The header should take up the entire width of the screen.
Image

9. These active effects should look different according to the Figma Mockup.
Image

KazymyrYan:
Hi! I've fixed all the issues from your review.

etojeDenys (changes requested):
almost done

Image

2. it would be useful to have the option to remove a product from the cart after pressing the button again

Image

Image

KazymyrYan:
Hi! I've fixed issues 2 and 3 from your latest review.

Regarding the 'Brand new models' section — after checking the data in public/api/products.json, I found that all products include both a full price and a discounted price, meaning there are no products without a discount in the dataset.

To handle this as best as possible, I've updated the filtering logic to:
1. Only include products from the most recent year
2. Sort them by discount amount (ascending), so products with the smallest discount appear first

Let me know if you'd like a different approach!

brespect (changes requested):
Hi! Thanks for the update and for checking the dataset.

For the **"Brand new models"** section, the discount logic actually shouldn’t be used. 
Even if the API contains both `price` and `fullPrice`, in this section we should display **only the regular (actual) price on the cards** and **not show the discount or the crossed-out price**.

So the filtering by discount isn’t needed here — just render the products and display the **current price only**.

Thanks!

KazymyrYan:
Hi! Fixed the 'Brand new models' section — products are now filtered by the most recent year only, and cards display the regular price without any discount or crossed-out price.

etojeDenys (approved):
great job

PR #933, opened 2026-03-11
add task solution

Abdulahhh2005:
[Demo Link](https://abdulahhh2005.github.io/react_phone-catalog/)


etojeDenys (changes requested):
good job

to improve:

1. these buttons should change photos

Image

2. add the option to remove the product from the cart by pressing the button again

Image

3. after selecting a product, the page should scroll upwards

Image

4. slightly decrease the size of the hovered image because it now looks merged with the other one

Image

5. page number should also be saved in the search parameters

Image


6. after changing 'items on page', the page number should be set to 1

Image

7. there is too much empty space

Image

etojeDenys (approved):
good job

last thing to improve:

1. that button should also remove the product

Image

PR #935, opened 2026-03-13
solution

OMedvid:
[DEMO](https://omedvid.github.io/react_phone-catalog/)


Denys-Kravchuk9988 (changes requested):
Good job!

1. It's better to add dropdown icons
Image

2. I would recommend to align these dropdowns with list of cards according to design
Image
Image

etojeDenys (changes requested):
good job

to improve:
1. it would be nice if the product could be removed by pressing the button again
Image

etojeDenys (approved):

PR #937, opened 2026-03-15
add task solutionv1

RomanPlyta:


etojeDenys (changes requested):
please provide a demo link and request the review again

RomanPlyta:
Sorry, I forgot. Here is the demo link. 
https://romanplyta.github.io/react_phone-catalog/#/

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. It's better to hide some tiles of pagination and leave not more than 8-10 tiles
Image
Like this, for example
Image

2. When I open phone info page I see this error
Image

etojeDenys (changes requested):
almost done

Image

2. it would be useful to be able to remove products from the cart by pressing the button again

Image

Image

3. the image is too big

Image

4. add a favicon and title to your tab

Image

etojeDenys (approved):
good job

to improve:
fix the colours because it does not look black
Image

PR #938, opened 2026-03-15
add task solution

serhiy23471:
**[Live Demo](https://serhiy23471.github.io/react_phone-catalog/)**


etojeDenys (changes requested):
great job, just a few fixes

1. add an address here because it currently leads to an error page

Image

2. it should be a link to the product page

Image

etojeDenys (approved):
good job

PR #941, opened 2026-03-17
Develop

DebHtm:
https://github.com/DebHtm/react_phone-catalog


brespect (changes requested):
Good progress, but try to pass all tests before PR submit

Anton-Kuchmasov (changes requested):
Oops, something went wrong with your deploy:

Image

Would you mind to double-check `"homepage"` in `package.json` file, re-check file paths for photos, make add-commit-push and then re-request review again?

If you're stuck, feel free to ask help in fe_chat. 

Also, your correct DEMO LINK is https://debhtm.github.io/react_phone-catalog/

etojeDenys (changes requested):
good job, let's improve your page

1. add some borders to the button

Image

2. it would be useful to be able to remove a product from the cart by pressing the button again

Image

3. whole card should be a link to the page, not just the image

Image

4. slightly decrease the hovered image scale because it looks merged with the others

Image

5. add a hover effect to the cards

Image

6. it should be 'all' instead of an empty space

Image

7. increase the price here as well
Image

8. make sure you remember to increase these numbers after pressing '+'

Image

9. after reloading the page, an error page appears for a couple of seconds. It would be better to display a loader

Image

brespect (changes requested):
Good progress, check next:
1. Seems like overlap with main page here: 
Image

2. Also, if you select something in "You may also like" section, page should smoothly scrolls to the top.

3. Add favicon and change browser tab title.

4. Pagination must be at the bootom of page, every time on every quantity of items:
Image

5. This button has no action:
Image

6. All links in Footer should be valid references to existing proper resources,
Image

2pasha (changes requested):
good job! 👏

here are some improvements: 

1. values here should be the same

Image

2. add border for button 

Image

3. this values should be selected for the product

Image

4. issue with scroll to top still exist: user click on some product in section "You may also like" -> page should be scrolled to top

Image

5. here should be relevant links

Image

6. user should have ability to visit product detail page by clicking on product in cart

Image

etojeDenys (changes requested):
almost done

1. after selecting a product, the page should scroll up

Image

2. it is better to open these links in a different tab

Image

vadiimvooo (approved):
Well done

PR #949, opened 2026-03-21
solution

Skay9n:
project: https://skay9n.github.io/react_phone-catalog/


etojeDenys (changes requested):
i really like the design, but let's improve the logic

Image

2. it would be useful to be able to remove products from the cart by pressing the button again

Image

3. fix these images

Image

4. that drop-down menu does not work
Image

5. make sure these links lead somewhere instead of an error page
Image

6. add a favicon and title

Skay9n:
i just removed discounted prices from brend new models because all new model in product.json have discount

brespect (changes requested):
Good progress, check next:
1. This hover effect should not overlap with the page background:
Image

2. When you select something in You may also like section, it should scrolls smoothly to the top of page:
Image

3. The bread crumb is not aligned correctly:
Image

4. Content of card should have same height every time, even when title is wide:
Image

etojeDenys (approved):
great job. last thing to improve

there is still a problem with the alignment. the icon is placed lower than the link
Image

PR #951, opened 2026-03-23
Develop

n7lots:
[DEMO-LINK](https://n7lots.github.io/react_phone-catalog/)


2pasha (changes requested):
good job! 👏

fix these moments: 

1. products don't load 

Image

2. issue with image

Image

3. change favicon and title

Image

4. fix linter errors and warnings

Image

brespect (changes requested):
Hello, some comments from previous mentors review were not fixed.

etojeDenys (changes requested):
i like that loader, but there are a few final fixes needed

Image

2. it would be useful to be able to remove a product by pressing the button again

Image

Image

3. these links should not lead to an error page

Image

Denys-Kravchuk9988 (changes requested):
Excellent work!

Tiny improvements:
1. It's better to use another color for this house icon on dark theme
Image

2. I would recommend to make color of this icon < consistent with text color
Image

brespect (changes requested):
Good progress, but let's pass all tests and linter checks, before requesting the review

brespect (approved):
Well done! 👍

PR #953, opened 2026-03-24
Develop

Stas-collab:
[DEMO LINK](https://Stas-collab.github.io/react_phone-catalog/)


etojeDenys (changes requested):
good job

to improve:

1. after changing the color, we noticed the problem

Image

2. cards are too wide

Image

3. there is too much space

Image

4. make sure that the product can be removed after pressing this button

Image

5. make sure these numbers increase as well

Image

6. this card should link to the product page

Image

7. these buttons should be wider

Image

Anton-Kuchmasov (changes requested):
Good job!

To improve:

1. Add some functional to `Checkout` buton (f.e., clear the Cart state and show notification e.g. "Thanks for your order!")

Image

2. Double-check UI styles of `Cart` page above - does they look so shrink on Figma mockup?

3. Footer links should be clickable. At least, add ref to your GitHub repo to the `Github` navlink

4. `Ipad 2020 4th gen` in `Rose Gold` color is not available. Would you mind to check again, how do you parse `color` search param from your URL?

Image

2pasha (changes requested):
good job! 👏

here are some improvements: 

1. change title and favicon

Image

2. please put points in the right place 

Image

3. in some product when we want to change color we receive this page, please double check available colors 

Image

4. make sure that you use correct fonts 

Image

etojeDenys (approved):
good job

PR #954, opened 2026-03-24
Develop

1umamaster:
[DEMO LINK](https://1umamaster.github.io/react_phone-catalog/#/)


etojeDenys (changes requested):
good job, to improve:

1. it would be nice if the product could be removed by pressing the button again

Image

2. make sure that these links do not lead to an error page

Image

3. after choosing that option, we see the page 'product not found'. you should only show the available colour and capacity options

Image

4. add a title and favicon

Image

Anton-Kuchmasov (approved):
Great job you did!



Before final review, I suggest you to add some easy functional on `Checkout` button click (e.g., erasing `Cart` state - that should be useful)

PR #955, opened 2026-03-25
Develop

zhyliaieva:
https://zhyliaieva.github.io/react_phone-catalog/


etojeDenys (changes requested):
good start

to improve:

Image

2. make your cards follow the Figma mockup
Image

example:

Image

3. It would be nice to remove product from the cart after pressing the button again

Image

4. there is too much space

Image

5. these buttons should work

Image

6. this section looks broken

Image

7. after selecting a product, page should scroll up

8. cards should not stretch

Image

9. the entire card should be clickable and link to the product page
Image

brespect (changes requested):
Good progress, check next:
1. When I click on Swiper buttons to see next image, this counter should works too:
Image

Image

2. When I pick something to favourites and check the header, this icon counters looks broken, they should have different positioning:
Image

3. Reduce the gap between cards, for now its too big, also I suggest to make cards wider:
Image

4. Make button Add to Cart/Added full width to the content of the card, as example you may check the Figma Mockup:
Image

5. Add some hover to all this images:
Image

6. Colors and capacity still not working, check also all previous comments:
Image

2pasha (changes requested):
good job! 👏

here are some improvements: 

1. remove this background on hover

Image

2. image should be centered 

Image

3. i would recommend to remove box-shadow on hover and leave only scale

Image

4. fix counters in header, example how to fix: 

Image

5. row gap is too large

Image

6. when user change some of these params we should show to him relevant product with that params 

Image

7. too big gaps in breadcrumbs

Image

8. do not forget to change counter when user change favorite items from favorite tab

Image

9. make sure that card looks exactly as in Figma

Image

10. it's great that user can open product details page on clicking anywhere here but user also should have ability to click on these icons and do not open product details page and do relevant action instead

Image

2pasha (changes requested):
good progress! 👏

here are last improvements: 

1. images still aren't centered 

Image

2. here is the bug when user click on some of capacity and app show to user another model (expected: iphone **13** 256gb, received: iphone **14** 256gb)

Image
Image

3. fix layout in the favorite page

Image

etojeDenys (changes requested):
1. these buttons stop change phone now

Image

2. there is too much space. footer should be visible when the cart is empty
Image

3. 'favorites' is empty, but it still shows 1 in the top-right corner
Image

4. it would be better if you made the images linkable as well

Image

5. slightly decrease the size of the hovered image to avoid merging with other items

Image

etojeDenys (changes requested):
last changes

1. make sure the footer is visible without scrolling when the 'favourites' section is empty

Image

2. fix the number of favourites here

Image

brespect (changes requested):
A few improvements:
1. I think this text should be at the same line:
Image

2. Better to choose default selection, now this picker looks strange as it merge with other colors and looks a bit broken, or add some gap, as you prefer:
Image

3. All links should refer to proper existing resourses, as example you may use Apple Inc Policy here, bad practise to redirect to the Not Found Page:
Image

etojeDenys (approved):
good job

PR #958, opened 2026-03-28
Done

VasyaRusnak:


etojeDenys (changes requested):
demo - https://vasyarusnak.github.io/react_phone-catalog/#/
1. the products in this section should not be discounted
Image

2. it would be nice if clicking the button again removed the product from the cart
Image

3. add hover effect to the images
Image

4. add hover effect to the product cards

5. make sure the cards are the same size

Image

6. you should save the sort values in the search parameters
Image

7. add some space here

Image

8. fix img here

Image

9. do not forget to implement this logic


Image

10. add a favicon and title to the page

Image

etojeDenys (changes requested):
1. selected color and capacity should have an active class

Image

2. these buttons do not work
Image

3. products in this section should not be discounted
Image

4. you should save the sort values in the search parameters

Image

Image

5. add the option to remove the product from the cart by pressing the button again

Image

Anton-Kuchmasov (approved):
Well done!

PR #960, opened 2026-03-31
Phone Catalog

YuliiaKosenchuk:
[DEMO LINK](https://YuliiaKosenchuk.github.io/react_phone-catalog)


brespect (changes requested):
Good progress, but your solution seems to be broken:
Image
Image
Image

brespect (changes requested):
Good progress, check next:
1. All this links must refer to existing proper resourses.
Image

2. When I select item in pagination, the page should not jump to top:
Image

3. Same way, when I select something in this section, it should smoothly scrolls page to the top:
Image

4. If you translate to the ukrainian language, the shortcut is "UA", "UK" - United Kingdom:
Image

5. Make this buttons smaller on Mobile View, for now they are bigger than links:
Image

2pasha (approved):
lgtm 🔥

P.S. remove unused code

PR #961, opened 2026-03-31
Solution

shatskov-artur:


brespect (changes requested):
Good progress, but you need to deploy and provide your DEMO LINK into Pull Request description, according to the requirement

shatskov-artur:
https://shatskov-artur.github.io/react_phone-catalog/

brespect (changes requested):
Good progress, check next:
1. Ad the favicon and change title for browser tab.

2. When I select something in this section, the page should smoothlty scrolls to the top, and also add some gap between this section and your Footer:
Image

3. Cards in Swiper should not be cropped:
Image

4. Pass all tests and linter checks before requesting the review.

PR #962, opened 2026-04-02
Develop

Evg3n22:
DEMO LINK(https://evg3n22.github.io/react_phone-catalog/)


brespect (changes requested):
Good progress, but you need to fix the tests and provide working DEMO LINK to the PR description

etojeDenys (changes requested):
almost done

1. image is too large, so scrolling appears
Image

2. don't forget to increase these numbers as well

Image

3. footer should not stretch

Image

4. font looks different
Image

5. add a title and favicon

Image

6. after reloading a non-homepage, an error appears

Image

Evg3n22:
If you find a problem in the cart page, delete `cart` from local storage

etojeDenys (changes requested):
demo link does not work. please fix it and request the review again

Image

etojeDenys (changes requested):
the demo has the same problem

 if you get stuck, feel free to ask for help in the chat

brespect (changes requested):
Good progress, check next:
1. Something went wron when I add to cart from product list and after that select that product and it should be already added, but I have the ability to add same product in Product Details Page one more time, so after that I got this screen:
Image

2. When you select something in "You may also like" section, it should smoothly scrolls to the top.

3. All Footer links should refer to existing resourses.

etojeDenys (changes requested):
almost done, just last fixes:

1. it would be better to open these links in a new tab

Image

2. recheck all your fonts because they look different

Image

Image

Denys-Kravchuk9988 (changes requested):
Good job! 

A few things to improve: 
1. White space appears on page where is the phone's info
Image

2. I would recommend to remove default button's border width
Image

3. Scroll should be disabled when menu is open
Image

brespect (changes requested):
Can't open your solution due to error:
Image

brespect (changes requested):
Seems like nothing changes from my previous review:
Image

etojeDenys (approved):
good job

PR #967, opened 2026-04-10
add task solution

PavloYurievich:


Denys-Kravchuk9988:
Checked this [link](https://pavloyurievich.github.io/react_phone-catalog/)

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. Please, remove default button's border-width
Image

2. I would recommend to add gaps between phones' cards
Image

3. When phone is added to cart is better to keep this white background and green text
Image

4. Tablets/Phones/Accessories aren't loaded when I open catalog's page
Image

5. I would recommend to export logos/icons in `.svg` format from Figma
Image
Image

2pasha (changes requested):
good job! 👏

here are some improvements: 

1. change favicon and title 

Image

2. it's better to add some real links in footer and open them in new page

Image

3. these numbers should be the same 

Image

4. user should have ability to open Product detail page by clicking on product in cart

Image

5. images in slider on mobile don't load

Image

6. fix images on phone (home page)

Image

etojeDenys (changes requested):
good job
to improve:

Image

2. now I can add the same product to the cart from both the home page and the product page. synchronise it

Image

3. these buttons should change the product

Image

4. after selecting a product from this section, the page should scroll up

Image

5. make this icon a link too

Image

6. there is too much space

Image

brespect (approved):
Lgtm 🚀

PR #969, opened 2026-04-15
Solution

pivkota:
https://react-phone-catalog-lovat.vercel.app/#/


etojeDenys (changes requested):
good job, to improve:

1. products in this page should not be discounted

Image

2. footer should not stretch
Image

3. you should save these values as search parameters

Image

Image

4. add favicon

Image

5. make sure these numbers are the same

Image

6. it should link to the product page

Image

brespect (changes requested):
Almost done, check this:
1. The should be some gap between pagination and footer.
2. Make pagination buttons a bit smaller:
Image

3. Definitely recommend to make this like ui select component:
Image

4. When I select something in You may also like section, I expect page will smoothly scrolls to top, now it is jumping:
Image

5. Some cards have different size on Mobile:
Image

etojeDenys (commented):
1. add a cursor pointer to this button
Image

etojeDenys (approved):
great job. just fix that comment

PR #972, opened 2026-04-16
react_phone-catalog

Bukavyna:
https://bukavyna.github.io/react_phone-catalog/


Bukavyna:
https://bukavyna.github.io/react_phone-catalog/

Denys-Kravchuk9988 (changes requested):
Very good job!

A few things to improve:
1. The image of empty cart wasn't loaded
Image

2. When I clicked on 8th page on phones and then switched to tablets list I see the empty list
Image
Image

3. By hover the card the top/bottom card's borders are hidden by wrapper container (the last card)
Image

4. Plus/Minus buttons are moved to left when price becomes high
Image

brespect (changes requested):
Good progress, check next:
1. When I select something in 'You may also like' section, the page should smoothly scrolls me to the top.

2. When I select pagination button item, the page should be at the same position.

3. Page should not blicking when I select filter, it looks strange:
Image

4. I change localisation to UA, but description on Product Details Page still in English, it looks broken for me as customer:
Image

5. Change the browser tab title and add the favicon.

Bukavyna:
То мені опис усіх товарів перекладати???

Denys-Kravchuk9988 (changes requested):
Good job!

One thing to improve:
1. I would recommend to not blick all page when User changes color of the phone or its size. It's better to add skeleton for the image and dynamic text (if it's dynamic) but leave colors and size buttons without blicking.
Image
Image

2. Here is needed to add favicon and appropriate title
Image

Denys-Kravchuk9988 (changes requested):
Excellent job!

One tiny improvement:
Please, add the loading state on image and disable filter's buttons when other phone is loading. Now the page is frozen by clicking the filter's button
Image

Anton-Kuchmasov (changes requested):
Hey, seems something is broken with `Price` styles here:

Image


Also, would you mind to slightly reduce empty `Cart` placeholder?
Let it have overall height `100vh` (both Header & Footer included here)

Image


When User is changing sorting & pagination params on Category page, changes should be added into URL. User might have an opportunity to share a link with friend, and all params should be saved:

Image

Take a quick look here ([reference](https://react-phone-catalog-lovat.vercel.app/#/phones?sort=alphabetically&perPage=8)): 

Image

Once again, your UI decisions and relization are perfect. Let's add some functional improvements! :)

vitalii-bondarenko-ua (approved):
Good job.

PR #973, opened 2026-04-17
Develop

redfield-mp:
[Demo link](https://redfield-mp.github.io/react_phone-catalog/#/)


brespect (changes requested):
Good progress, check this:
1. Contact and Rights pages should not be empty, should link to proper existing resourses or well implemented own pages:
Image

2. Remove broken items:
Image

3. All card images should be same size:
Image

4. Products in this section should not have discount:
Image

5. When you select something in "You may also like" section, page should smoothly scroll to the top.

redfield-mp:
1. Added content.
2. Fixed.
3. Different images have different sizes. Should I crop images?
4. If the product should not have a discount, then what type of cost should I show: "priceRegular" or "priceDiscount"? If I remove the discount from the "Brand new models" section, then should I remove it from the "ProductDetailsPage"?
5. Added.

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. By change color the phones images should also be changed to appropriate color (same for memory size)
Image

2. I would recommend to add logic when user clicks on Added to cart button the phone is deleted from cart.
Image

3. By change the count it's better to recalculate sum
Image

4. Phones in this section shouldn't be with discount according to design
Image
Image

5. It's better to align footer's content to left side according to design
Image
Image

redfield-mp:
1. Done.
2. I need some clarification on this task, as it doesn't seem to be mentioned in README.md. 
3. Done. 
4. Done. 
5. Done.

vadiimvooo (changes requested):
Almost done. Just fix add to cart button to be able to remove item from cart as Denys suggested in previous comment. The behaviour should be the same as with favourite button. Please feel free to ask for help in fe_chat.

redfield-mp:
Done. But I have the following suggestion in the README.md file:

> 2. If the product is already in the `Cart` the button should say `Added to cart` and do nothing.

vadiimvooo (approved):
Well done

PR #976, opened 2026-04-23
add task solution

IvanovvvIhor:


2pasha (changes requested):
1. change title and favicon 

Image

2. items in this section should be without any discount 

Image

3. make sure that you are using right fonts 

4. make sure that you have done design from Figma here

Image

5. add real links here

Image

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. When I choose 4 cards per page the arrow buttons are hidden
Image

2. It's better to disable scrolling when menu is open
Image

3. I would recommend to export icons/logos in `.svg` format from Figma for better quality
Image
Image
Image

4. Breadcrumbs go outside the phone size
Image
Image

5. Checkout block is broken on phone
Image

etojeDenys (changes requested):
good job, to improve:

1. loader looks broken

Image

Image

Image

3. it would be better to open this link in a new tab

Image

4. add some space here

Image

2pasha (changes requested):
great job 🔥

please fix last moments: 

1. footer always should be in the page bottom 

Image

2. items here should be without discount 

Image

Anton-Kuchmasov (approved):
Great job you did!

PR #981, opened 2026-04-27
Develop

berezandiana:
https://berezandiana.github.io/react_phone-catalog/


brespect (changes requested):
Good progress, but you need to pass all tests and linter checks before requesting mentor's review

brespect (changes requested):
Almost done, cehck this:
1. When you select page in pagination section the page should not scrolls to top:
Image

2. You have localization for your app, its really nice thing, but it is strange for me that this section is on English while I select the UA:
Image

3. When I select something in "You may also like section" I expect the page will scroll to the top:
Image

4. "Back" button under the "Bread Crumbs" should back me to the navigation, as example to toablet, not to previous model I selected:
Image

5. All links in Footer should refer to proper existing resourses.

6. Remove vertical scroll on empty page:
Image

7. Make the logo default black color on Mobile, as for now, it is hard to see it in side menu:
Image

Anton-Kuchmasov (changes requested):
Good job!

To improve:

1. Remove this bars (both):

Image

2. It's better to reset `[Css Autofill](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:autofill)` styles in your Search bar: 

Image

3. On some accessoires' pages, first color of the list is not clickable. Is it `space gray`? It should work too

Image

etojeDenys (changes requested):
last fixes:
1. products in this section should now be discounted

Image

2. this card should link to the product page

Image

3. slightly decrease scale of the hovered image to avoid merging and overlapping

Image

4. it would be useful to be able to remove the product after pressing the button again

Image

Denys-Kravchuk9988 (approved):
Excellent job!

PR #983, opened 2026-04-29
Mit

mitnues:


IanLuan (changes requested):
Your code looks good, but the DEMO LINK is missing. Please run the deploy.

mitnues:
"https://mitnues.github.io/react_phone-catalog/"

pedro-ruas (changes requested):
The design of the page doesn't match the one proposed by the task.
- Fonts are different
- Items positioning is different
- Colors are different
- Multiple functionalities do not work
  - Products do not load in any page
  - Icons don't change with interactions
  - Images do not load
 - There are also elements that should not be there, like the reload button in catalog pages

mitnues:
it was working good in npm start, mabe e had problem in demolink again.

mitnues:
https://mitnues.github.io/react_phone-catalog/

IanLuan (changes requested):
Great progress, you’re almost there! However, there are a few changes needed to improve your project:

- The design could be closer to the mockup. Review the font family, the **Shop by Category** section, the cart and favorites pages, the product page, and the home page carousel.
- The product does not change when selecting another color or capacity.
- The pagination has a bug: if I select 4 items per page, it displays 4 items, but when I go to the next page, it shows many items.
- The search is not working properly. As soon as I start typing, it freezes the page while searching. If you want to keep it, add debounce to the search input.

IanLuan (changes requested):
Great progress! You solved almost all the problems. However, when I change the capacity of a product inside the product details page, the product does not change. It should work the same way as the color selection.

IanLuan (approved):

PR #985, opened 2026-05-02
Develop

OwlDevcUA:
[Link] - https://owldevcua.github.io/react_phone-catalog/


Anton-Kuchmasov (changes requested):
Good work!

To improve:

1. This card shouldn't stretch - even if there is only 1 card in `Favourites` section at the moment:

Image

2. 16 items on page looks like only 4. Would you mind to fix this?

Image

3. Add links to your Github, Linkedin etc. at the `Footer` navlinks

etojeDenys (changes requested):
almost done

Image

2. it is better to open these links in a new tab
Image

3. make sure you are increasing these numbers as well

Image

etojeDenys (approved):
good job

PR #986, opened 2026-05-04
solution

u5135039754-dev:
[Demo](https://u5135039754-dev.github.io/react_phone-catalog)


Anton-Kuchmasov (changes requested):
There are a few critical deploy' issues need tp be fixed. Please re-deploy your App, double-check if everything is OK and then re-request review again

thread in fe_chat: https://mate.academy/chats/20607?selectedMessageId=11281255

brespect (changes requested):
Good progress, but for now i can't open your DEMO LINK, also make sure that you pass all tests before requesting next review

etojeDenys (changes requested):
good job, to improve:

1. these buttons should work

Image

2. instead this part loader should appear

Image

3. drop-downs should close after an option has been selected or when the user clicks outside the container

Image

4. these filters should be saved as search parameters

Image

Image

5. make sure that these buttons change the product

Image

brespect (changes requested):
Good progress, check next:
1. When I select something in pagination, the page should not scrolls to the top:
Image

2. All links in Footer should refer to proper existing resourses.

3. Add the favicon to browser tab.

4. The slider should works with "Order Now" button:
Image

5. When I select something in "You may also like" section, the page should smoothly scrolls up:
Image

Anton-Kuchmasov (changes requested):
Great work you did!

To improve:

1. Add some functional to `Checkout` button. It's clear that our web-app doesn't have a full-scaled Cart functional, but you can show just a small clarification window (e.g. `Thanks for you order!`) and clear the Cart state:

Image

Just a link that can be used as a reference: https://owldevcua.github.io/react_phone-catalog/

2. Would you mind to make slider (basically, ALL your page sliders) smoother? See reference link to clarify

3. Add another slides to Main slider - now it seems to be slightly bored

brespect (changes requested):
Good progress, check next:
1. When I select something in this scetion, the page should scrolls up. but got not found instead:
Image

2. Not found Page should has Header and Footer as other pages.
Image

3. Don't show the counters when there is nothing in Favourites/Cart:
Image

4. Don't use same image three times in slider,  let's add another one:
Image

5. Let's reduce the size of font on Mobile screen:
Image

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve: 
1. The slider doesn't reset timer when I click on sticks. When I open second slide wait 3-4 seconds and then press prev stick the first slide is immediately switched to second slide.
Image
Image

2. I would recommend to add`cursor: pointer` by hover on interactive image
Image

3. It's better to add some loading state for this page and when color and configurations are changed
Image

4. Plus/Minus buttons are better to be in a column
Image

5. I would recommend to disable scrolling when menu is open (on main page it's already disabled)

6. In cart I have two products but I don't see any badge here
Image

2pasha (approved):
lgtm 🔥

PR #989, opened 2026-05-06
solution react_phone-catalog

opalahecha:
[DEMO](https://opalahecha.github.io/react_phone-catalog/)


brespect (changes requested):
Good progress, almost done check next:
1. When I search "Iphone Gold" I want to get all golden phones that are presented, including all models. For now it shows me nothing.

2. Back button should back me to the previous layer, not to the previous Iphone. In this case it should be Phones Catalog Page:
Image

3. All links in Footer should lead to proper existinmg resources.

4. There should be no scroll on empty page, check the shopping cart also:
Image

5. Search Bar should not be cropped on small mobile screeen:
Image

brespect (changes requested):
Still cropped:
Image

2pasha (approved):
lgtm 🔥

PR #991, opened 2026-05-07
add solution

dimon2202:
[DEMO LINK](https://dimon2202.github.io/react_phone-catalog/)


2pasha (changes requested):
good job! 👏

here are some improvements: 

1. change title and favicon 

Image

2. items here should be without any discount 

Image

3. add some hover effects for these items 

Image

4. it's better to add `cursor: pointer` for all your interactive elements 

Image

5. back button should navigate user one step back, right now it is navigate to home page

Image

6. numbers in both places should be the same 

Image

7. add some message after clicking on checkout button 

Image

8. this button doesn't work 

Image

9. it's better to add there some external links which will be opened in new tabs 

Image

Anton-Kuchmasov (approved):
Awesome!

PR #996, opened 2026-05-17
done

bcheban:
[DEMO](https://bcheban.github.io/react_phone-catalog/)


brespect (changes requested):
1. Search should show me all Purple Phones when I typed: "Iphone Purple", but for now nothing:
Image

2. If I select something in "You may also like" section, the page should smoothly scrolls to the top:
Image

3. Same way if I select next previous page, the page should not:
Image

4. If you have localization, you need to translate everything, as it looks strange:
Image

Anton-Kuchmasov (commented):
Awesome!

Anton-Kuchmasov (approved):

PR #1003, opened 2026-05-30
 react_phone-catalog_solution

Banderos14:
https://banderos14.github.io/react_phone-catalog


Denys-Kravchuk9988 (changes requested):
Good job!

Checked this link
https://banderos14.github.io/react_phone-catalog

A few things to improve:
1. It's better to use HashRouter instead of BrowserRouter because the BrowserRouter doesn't work appropriately with GH Pages.
Image

2. Pages don't load their content. Api links don't work.
Image

3. After pressing `Reload` button the 404 page is open
Image

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve: 
1. These configurations should also change slug in url
Image

2. Header looks broken on phones. I would recommend to add some burger icon
Image

3. It's better to implement select with params `Items per page` and implement pagination.
Image

4. On favourites page cards took all page's place. When there is >4 cards it looks good
Image
Image

5. Plus/Minus icons are better to be centred
Image

6. Slider is better to open next each next slide every 4-5 seconds
Image

vitalii-bondarenko-ua (changes requested):
Good job. Almost done.
- in product page active filters should match product in url
- instead disable product which already exists in cart make button active but with text and functionlality remove from cart.

Image

- brokent image in cart page
- center content in minus/plus buttons

Image

- add pagination and per page dropdown in product list pages

brespect (changes requested):
Good progress, check next:
1. Products in this section should not have discounts:
Image

2. I select two Apple Watch Series items and want to search them in Favourites, but search term "Apple Series" does not work:
Image

3. Add the favicon.

4. Back button should not me back to the previous phone, it should back me by layer, to Phones Page in this case:
Image

5. When you select something in this section, the page should smoothly scrolls to the top:
Image

6. Implement the ability to select how many items user wants to see at phones. tablets, accessories pages and the pagination.

etojeDenys (changes requested):
to improve:

1. make sure you fix all the problems from the previous review

2. this card should link to the product detail page
Image

3. the images are too big

Image

4. these links do not work

Image

brespect (changes requested):
My previous comments was not fixed, especially 6 point, also check this:
Image
When I select something in this section it routes me to Not Found Page

Banderos14:
> My previous comments was not fixed, especially 6 point, also check this: Image When I select something in this section it routes me to Not Found Page

In my Figma design, this section is not implemented as shown in your screenshot. The original design only includes category cards with images on the homepage (Phones, Tablets, Accessories). When the user clicks on a category card, it should directly open the corresponding category page.
The version with the three text buttons is not part of the Figma mockup, so it was not intended in the original design.

brespect (changes requested):
Almost done:
1. Align text to left, and add some padding for arrow icon:
Image

2. All links should refer to existing resources:
Image

3. Add the favicon and change title for browser tab.

brespect (approved):
Lgtm

Banderos14:
> Lgtm

I just found a problem with incorrect links to phone numbers, and I also messed up the photos—they're in the wrong sections instead of where they're supposed to be according to the description. I'll fix it right away.

PR #1005, opened 2026-06-01
https://github.com/InhaKr/react_phone-catalog/tree/dev

InhaKr:


etojeDenys (changes requested):
demo link does not work. please fix it and re-request the review

InhaKr:
https://inhakr.github.io/react_phone-catalog/

etojeDenys (changes requested):
good job, to improve:
1. slightly decrease the size of the hovered images to avoid merging and overlapping

Image

2. these buttons should work

Image

3. looks broken

Image

4. after selecting a product from this section, the page should scroll up

Image

5. font looks different

Image

6. these parameters should be saved as search parameters so that they are kept after the page is refreshed

Image

Image

7. it is better to open this link in a new tab
Image

8. footer should not stretch
Image

PR #1007, opened 2026-06-02
Develop

VladChudin:
In the future, I also plan to add a field for detecting and debounce.
[[DEMO LINK](https://react-phone-catalog-1-ymxp.onrender.com)]


brespect (changes requested):
Good progress, but you need to pass your working DEMO LINK to the PR description and also pass all CI checks (including linter) before reqiesting the next review

VladChudin:
Could you please help with the demo link, I tried to fix it but it didn't work.

brespect (changes requested):
Answered you in fe_chat

VladChudin:
Locally the linter doesn't complain, but in tests it complains about things I've already fixed.

brespect (changes requested):
Good progress, check next:
1. This message appears when I select item in you may also like section:
Image

2. Also Footer should be every time at the bottom of the page.

3. Add the favicon and change title for browser tab.

4. All this buttons should work and refer to all existing resourses:
Image

5. Add hover to all interactive elements: 
Image

brespect (changes requested):
Few things that I notice:
1. Page is blicking when I select product params, it should not:
Image

2. Pagination should not scroll user to the top:
Image

3. But You may also like section should:
Image

4. Remove this empty space between right border and cart & favourites icons, check how it looks on Figma mockup:
Image

VladChudin:
Regarding points 2 and 3, I don’t think this is a necessary step. The task doesn’t mention anything about it, and this is how most online shops operate, so I’d like to leave everything as it is. As for the ‘You may also like’ section, I also believe it’s displaying the correct products; I’m not quite sure what exactly the issue is. I’ll fix points 1 and 4 now

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. After reloading of page I see 404 error.
Image

2. Mistype on breadcrumbs
Image

3. There aren't icons of product's count
Image

4. I would recommend to adjust +/- buttons by one column
Image

5. It's better to implement the feature by second click on button should remove product from cart
Image

VladChudin:
Regarding the feature that removes an item from the basket when clicked again: I had already implemented this, but my mentor told me to remove it, so I did. As for aligning the buttons in a single column, I don’t see the point—when the price is too long, it pushes the item itself out of alignment, making it look uneven. As for the product count icon, I removed it because it wasn’t aligning properly; there are very complex calculations involved for the indentation. If you know a simpler way to do it, please let me know. I’ll fix the rest now.

etojeDenys (changes requested):
to improve:

1. after reloading, an error occurs on a non-homepage
Image

Image

3. this card should link to the product details page

Image

Anton-Kuchmasov (changes requested):
Almost done!


When I'm trying to reach into product card from Cart, App redirects to 404 :(

would you ming to fix it?

Image

Image

etojeDenys (changes requested):
great, last two fixes:
1. image is too big, so users have to scroll down to see the footer

Image

2. make these photos into links as well

Image

VladChudin:
I've fixed the links and made the image a bit smaller, but the indents are absolutely necessary, and even if the footer isn't visible, it's not a problem. Can you please accept the assignment?

VladChudin:
And it's not the image that's the problem. The problem is that the box takes up 100vh so the footer is pressed to the bottom. My mentor told me so, so it's not my fault.

etojeDenys (approved):
great job!

PR #1010, opened 2026-06-05
Develop

RuslanV23:
https://ruslanv23.github.io/react_phone-catalog


RuslanV23:
I started from a blank project, so the Mate scripts were lost. I hope you'll still accept the assignment.

brespect (changes requested):
Almost done, but you need to improve this:
1. Let's remove scroll on empty page:
Image

2. Pass all tests before requesting the next review.

RuslanV23:
> Almost done, but you need to improve this:
> 
> 1. Let's remove scroll on empty page:
> 
> Image
> 2. Pass all tests before requesting the next review.


I fixed the empty cart issue.

 However, I can't run the Mate tests because they were probably lost when I created the project from scratch. What should I do? Should I try to set up those scripts again?

brespect (changes requested):
Good progress, but it is better to pass all tests, you may sync with base branch using UI, or do it through terminal, if you need some help better to ask in fe_chat

RuslanV23:
Added and passed tests😎

Denys-Kravchuk9988 (changes requested):
Excellent job!

A few things to improve:
1. The images' sizes are better to be equal
Image

On design:
Image

2. Some content goes outside on the phone/tablet screen size
Image

3. On tablet the adaptation seems to be broken
Image

4. I would recommend to reduce paddings on tablet
Image

Anton-Kuchmasov (approved):
Awesome!

PR #1011, opened 2026-06-06
Develop

AlexanderBarkar:
Здраствуйте)Наконец-то я его сделал,спустя больше месяца...Должно быть все идеально ,перепроверял 10 раз что бы в соотвецтвии с фигмой и все работало,не судите строго)


AlexanderBarkar:
https://alexanderbarkar.github.io/react_phone-catalog/#/

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. `Brand new models` and `Hot prices` are better to be on the same column according to design
Image
Image

2. When cart is empty it's better to attach footer to the bottom
Image

the same for favourites
Image

3. It seems that this paddings are odd. The cards' width is less than on design
Image

4. The logo and icons aren't loaded on phone screen size
Image

5. It seems that the adaptation is broken on phone/tablet
Image
Image
Image
Image

6. I would recommend to implement slider on phone screen size as on design
Image

7. Logo isn't loaded here
Image

8. It's better to implement cart design on tablet/phone screen size as on design
Image
Image

AlexanderBarkar:
Постарался с помощью менторов исправить все косяки.Скажите если еще что-то есть то опишите подробно как поправить пожалуйста.

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve: 
1. By clicking on silver color I receive `Product not found`. Try to debug this link: https://alexanderbarkar.github.io/react_phone-catalog/#/accessories/apple-watch-series-3-42mm-42mm-silver
Image
Image

2. Charasteristics are better to be aligned with design
Image
Image

3. Images are better to have equal size
Image

4. These icons on phone should be aligned to center and it needs to add count badges.
Image

5. It's better to disable scrolling when menu is open

6. Select should have margin from right side of screen
Image

7. Pagination is better to be in a row on phone
Image

8. Here is better to have margins also on tablet
Image

etojeDenys (changes requested):
good job, to improve:

1. products in this section should now be discounted

Image

2. this link should open in a new tab

Image

3. these values should be saved in the search parameters so that they are kept after the page is refreshed

Image

Image

4. if I select a different colour or capacity, I see a 'Page not found'
Image

5. make sure these numbers are equal

Image

6. it should link to the product detail page

Image

7. add a favicon and title to the page

Image

Denys-Kravchuk9988 (approved):
Looks good to me!

PR #1012, opened 2026-06-11
React Phone Catalog

VitaliyHoroshko:
https://vitaliyhoroshko.github.io/react_phone-catalog/


brespect (changes requested):
Good progress, cehck next:
1. There should be no discount for new models:
Image

2. If I select something in this section it should smoothly scrolls to the top, not reload the page:
Image

3. Use Apple Privacy & Policy here;
Image

4. Back button should go back buy layer, in this case to the phones, not to the previous one:
Image

5. Don't forget to pass all tests before requesting next review.

brespect (approved):
Well done!

PR #1014, opened 2026-06-13
Develop

ValentynaITCh:
[****](https://valentynaitch.github.io/react_phone-catalog/)


Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. It's better to improve pagination. By clicking on prev and next arrows the page should be decreased or increased. Active pagination's tile should always be visible.
Image

2. Page's content goes outside on phone's screen size
Image

3. It's better to disable scrolling when menu is open
Image

4. By clicking on arrows the card is partly hidden
Image

5. I would recommend to replace symbols on icons (export in `.svg` from Figma)
Image

vadiimvooo (changes requested):
Good job, a few things to improve:
1. It is better to have the same padding for all sections (on tablet size)
Image

2. Looks like bottom padding is too big
Image

3. Please make it possible to remove item from cart when you click on the button
Image

4. The placement of sections in the Product Item should be fixed. Please refer to the mockup.
Image
Image

brespect (changes requested):
1. When I try to Order Now I got same phone every time (Iphone 14 pro) and Not Found page every try:
Image

2. This links should refer to proper existing resources:
Image

3. When I select item in Pagination it should not scrolls up the page after:
Image

4. Back button should refer me back the by layer, not the previous tablet in this case:
Image

brespect (approved):
Well done!

PR #1016, opened 2026-06-14
Added solution

MykolaFatkullin:
https://mykolafatkullin.github.io/online_store/#/tablets


brespect (changes requested):
Good progress, check next:
1. Add favicon and change browser tab title.
2. There should be no discounts in this section:
Image

3. Hover effect should not overflow another image:
Image

4. This links should have different gap between them and should refer to proper existing resources:
Image

5. The images in the slider are really great, but they look flattened, fix that:
Image

6. If I select something in this section it should smoothly scrolls to the top without rerender:
Image

7. Back button should back me not to previous Iphone, but to previous layer to Phones Catalog in this case:
Image

8. Footer should be sticky at the bottom of the page:
Image

9. The siz e of this sections should not jump when I increase or decrease product count:
Image

10. Pass all tests before requesting the next review.

MykolaFatkullin:
2. If I remove all products with no discount there will be no products. As on JSON file we have only products with discounts.

brespect (changes requested):
Good progress, but for products with discounts in Brand New Model section you can implement the flag that can turn on/off discount visibility:
Image

etojeDenys (approved):
lgtm

PR #1017, opened 2026-06-15
Add solution

Yevhen-Srdk:
-[DEMO LINK](https://yevhen-srdk.github.io/react_phone-catalog/)


2pasha (changes requested):
good job! 👏

here are some improvements: 

1. change favicon and title

Image

2. it will be great if user can open product details page by clicking here 

Image

3. on phone slider doesn't work 

Image

brespect (changes requested):
Good progress, check next:
1. There should be no horizontal scroll and probably image should not overflow the Footer:
Image

2. Pagination Item after selection should not scroll up the page.

3. Pagination should be at the bottom of the page every time, no matter how many items are shown:
Image

4. This button do nothing:
Image

5. All links in Footer should lead to proper existing resources:
Image

6. Back button should back me by layer, not to the previous phone/tablet/accs, but to Phones/Tablets/Accessories Catalog:
Image

brespect (approved):
Well done!

PR #1019, opened 2026-06-18
success

KKRuslan:
[DEMO LINK](https://KKRuslan.github.io/react_phone-catalog/)


Denys-Kravchuk9988 (changes requested):
Good job!

1. The page of some phone doesn't work
Image

2. Icons are not loaded
Image
Image

3. By hover on the card the part of phone hides on top and bottom
Image

4. I would recommend to remove the item from cart by clicking on `Added to cart` button
Image

5. On tablet the adaptation looks broken
Image

6. Some part of content goes outside the screen
Image

7. It's better to disable scroll when menu is open
Image

8. The icon on phone's menu shouldn't have any border around icon
Image

etojeDenys (changes requested):
good job, to improve:
1. it should not be empty
Image

Image

3. products in this section should not be discounted

Image

4. it seems that you forgot to add this section

Image

etojeDenys (approved):
lgtm

PR #1020, opened 2026-06-18
add task solution

annaskobara:
https://annaskobara.github.io/react_phone-catalog/


Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. It's better to increase card's size on desktop according to the design
Image
Image

2. The timer is not resetted by clicking on slider's arrows. So, when I wait 5-6 seconds and click on next/prev arrow the slider just opens other slide immediately. It's better to reset timer when user clicks on arrows
Image

3. Please, re-check the font on whole website. It seems that some text has default font 
Image

4. Close icon isn't loaded
Image

5. The slider doesn't show the last phone's card
Image

6. The adaptation on pages Phones/Tablets/Accessories looks broken
Image

7. The logo should also be visible on menu
Image
Image

brespect (changes requested):
Good progress. check next:
1. There should be no discounts on brand new model items:
Image

2. Check the gap between items on wide screen:
Image

3. If I select something in `You may also like` section the page should smoothly scrolls to the top.

4. Back button should redirect me back by layer, not to the previous phone/tablet but to the catalog:
Image

5. This button do nothing for now:
Image

6. This page better to be fullwidth not to the right side:
Image

etojeDenys (changes requested):
good job, to improve:
Image

2. the image should also link to the product detail page

Image

3. it looks broken. on the desktop version, there should only be four cards in one row

Image

Anton-Kuchmasov (changes requested):
Something went wrong with PC' view layout:

Image

Please fix it, double-check is everything works great, don't forget to make new deploy and then re-request review again

etojeDenys (approved):
good job

PR #1021, opened 2026-06-20
Finally

tavorotuk:
https://tavorotuk.github.io/react_phone-catalog/


etojeDenys (changes requested):
great job, to improve:
1. increase the price here as well
Image

Image

3. add a title and favicon to the page

Image

Denys-Kravchuk9988 (changes requested):
Good job!

One thing to improve:
1. The adaptation on tablet looks broken
Image
Image

tavorotuk:
> Good job!
> 
> One thing to improve:
> 
> 1. The adaptation on tablet looks broken
> 
> Image Image

That’s because I set a fixed width and height for each Product Card for every possible screen size. As a result, the Product List shows only as many cards in a grid as can fit horizontally on the current viewport.

If the screen is set to 768 pixels, as shown in the figma-design, three cards will fit perfectly. If it is one pixel less, two cards will fit.

Please suggest how to solve this problem. 

Знімок екрана 2026-06-21 о 16 35 18
Знімок екрана 2026-06-21 о 16 36 37

etojeDenys (approved):
lgtm

PR #1029, opened 2026-06-26
Develop

Y-Kotvitskiy:
Pleace, check my app


Denys-Kravchuk9988 (changes requested):
Checked this link. It seems that the project wasn't deployed properly
https://y-kotvitskiy.github.io/react_phone-catalog/

etojeDenys (changes requested):
great job, to improve:
1. these buttons should change the photos
Image

Image

3. these numbers should be the same

Image

4. increase the price here as well

Image

5. after reloading the homepage, an error occurs
Image

6. it would be nice if the product could be removed by pressing this button again

Image

Y-Kotvitskiy:
Corrected all comments made in the review

brespect (changes requested):
Good progress, check next:
1. Order now button do nothing:
Image

2. Add the favicon.

3. Better to move this settings buttons to header next to the favourites icon:
Image

4. Slider buttons looks not good on white screen:
Image

5. Header should be full page widht amd have underline:
Image

6. When I select something in this section page should smoothly scrolls to the top:
Image

etojeDenys (approved):
good job

PR #1030, opened 2026-06-27
react_phone-catalog

valerij0412:
https://react-phone-catalog-ie3w-rac4x8des.vercel.app/


etojeDenys (changes requested):
1. add some space to the whole page on both sides
Image

2. these buttons look almost the same. give the button a different background color if the card is added

Image

3. change text color here

Image

Image

4. add a favicon and a title to the page

Image

5. add some space before the footer

Image

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve: 
1. Breadcrumbs lead to unexisting page
Image

2. I would recommend to limit content size by 1200px and centre it by `margin: auto`. 
Image
Image

3. It's better to implement badges as on design
Image
Image

4. It's needed to add favicon to the page
Image

5. Here should also be badges
Image

6. The adaptation on phones looks broken. Some content goes outside the page
Image

7. White line under menu point should be on the bottom header's line
Image
Image

8. I would recommend to remove border from cart's block and +/- buttons should have equal width/height
Image

2pasha (changes requested):
no access to your DEMO LINK

Image

etojeDenys (changes requested):
almost done

1. selected page numbers should have 'active' style

Image

Image

3. there is too much free space here

Image

4. it would be useful to be able to remove a product from the cart by pressing the button again

Image

2pasha (approved):
lgtm 🔥

PR #1034, opened 2026-07-07
The portfolio project

tavokina:
[DEMO LINK](https://reactphone-catalog-rust.vercel.app/)


brespect (changes requested):
Good progress, but you need to pass all tests and share working DEMO LINK to the PR description before requesting the review

brespect (changes requested):
So, you need to share access everyone who wants to watch your DEMO, without requesting it
Image

2pasha (changes requested):
good job! 👏

here are some improvements: 

1. change title and favicon 

Image

2. add `cursor: pointer` for all interactive elements

Image

3. add hover effects for these elements 

Image

4. no need to show all pages 

Image

5. i have added only grey color to card, but other colors also looks like in card (same with capacity)

Image
Image

6. it would be great to open product details page by clicking here

Image

7. this image looks too large 

Image

brespect (changes requested):
Good progress, check next:
1. Pagination buttons should be sticky at the bottom of the page:
Image

2. After I select item from You may also like section page should smoothly scrolls to the top, not reload:
Image

3. Back button should back me by layer not to the previous item but to catalog:
Image

4. "Apple Yellow" gives me no results but there are a lot of yellow phones:
Image

5. Make search same height as favourite and cart icons.

6. Footer looks broken on wide screen:
Image

7. All links in Footer should lead to proper existing resources.

2pasha (changes requested):
great progress! 🚀

last thing: let's style this input to match entire design system (minimum is to increase input height)

Image

Zoolgrand (approved):
All issues have been fixed. Great work
Approved

PR #1035, opened 2026-07-10
first version of the Phone catalog project

SerhiyShimko:


etojeDenys (changes requested):
the demo appears to be broken. please fix it and re-request the review
image

SerhiyShimko:
Hello! This is an updated version of the project, I fixed the problem with the URL and now everything should work correctly https://serhiyshimko.github.io/react_phone-catalog/

Anton-Kuchmasov (changes requested):
Good job!

To improve:

1. Link `Rights` in the footer doesn't work as expected
2. Would you mind to normalize `Cart` numbers? F.e., this 3 numbers (see screenshot) should be consistent

Image

Denys-Kravchuk9988 (changes requested):
Good job! 

A few things to improve:
1. The slider should always show 4 cards in row. So, next button should be disabled when the last card is shown on right side of slider
Image

2. When I press manually on slider buttons the timer is better to be resetted. Now, when I wait 3-5 seconds and press next button the two times slider is executed
Image

3. On ukrainian version the breadcrumbs aren't translated
Image

4. Page looks broken when breadcrumbs are very long on phone (https://serhiyshimko.github.io/react_phone-catalog/#/product/apple-watch-series-6-40mm-silver)
Image

5. Here also should be added `models` according to design (34 models for example)
Image

6. I would recommend to add after ... also the number of the last page
Image

7. I would recommend to add logic that by second click the product is removed from cart
Image

8. On tablet (768px) should be 3 cards in row according to design
Image

SerhiyShimko:
Hi, I made all the edits

brespect (changes requested):
Good progress, check next:
1. When I click on Order Now button it do nothing:
Image

2. When I select something in You may also like section page should smoothly scrolls to the top:
Image

3. Add the favicon.

4. Give me the ability to click anywhere I want in this area not just on slider point:
Image

5. If you have localisation in your App, you neec to translate all elements:
Image

6. There are a lot of empty space. So "+" and "-" buttons should not be in the next line:
Image

7. Pagination should be sticky at the bottom of the page:
Image

8. If I changed the tab, as example from tablet to phones, pagination filter should not reset from 16 or all to 4.

9. So, we`ve got Yellow Apple Iphones, but the search gives me nothing:
Image

10. Try to pass all CI checks, especially Linter, before requesting a review.

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. I would recommend to reduce this image and move it to center
Image

2. Plus/Minus buttons should be on the same vertical line
Image

3. It's better to add some space between title and breadcrumbs
Image

4. One slide on slider hasn't loaded a picture
Image

etojeDenys (changes requested):
almsot done, last fixes:
1. it should also link to a product detail page

Image

2. make the photo smaller so that the footer is visible without scrolling

Image

roman-mirzoian (approved):
Well done 👍

PR #1037, opened 2026-07-13
add solution

boggeniys:
**[DEMO LINK]** https://boggeniys.github.io/react_phone-catalog/


etojeDenys (changes requested):
demo does not work. please fix it and re-request the review

brespect (changes requested):
Good progress, check next:
1. Order Now button do nothing right now:
Image

2. Add favicon and correct title to browser tab.

3. Products in this section should have no discounts:
Image

4. When you select something in this section page should not reload, but scrolls smoothly to the top:
Image

5. Back button should back me by layer to catalog and not to previous phone/tablet:
Image

Image

etojeDenys (changes requested):
great job, just a few last fixes:
1. make sure these numbers are the same

Image

2. it would be useful to be able to remove a product from the cart by pressing the button again

Image

roman-mirzoian (approved):
Well done 👍

PR #1038, opened 2026-07-13
Solution

nyarachun:
[DEMO LINK](https://nyarachun.github.io/react_phone-catalog/#/)


brespect (changes requested):
Good progress, check next:
1. Order Now button do nothing right now:
Image

2. Add favicon and to browser tab.

3. Products in this section should have no discounts:
Image

4. When you select something in this section page should not reload, but scrolls smoothly to the top:
Image

5. Back button should back me by layer to catalog and not to previous phone/tablet:
Image

Image

7. "Apple Yellow" search gives me no results, but must return at least yellow phones:
Image

Anton-Kuchmasov (approved):
Good job!

PR #1039, opened 2026-07-14
Solution

Rostyslav452:


Rostyslav452:
DEMOLINK https://rostyslav452.github.io/react_phone-catalog/

brespect (changes requested):
Good progress, check next:
1. Something went wrong with details page:
Image

2. When I select something in You may also like section page should smoothly scrolls to the top:
Image

3. Back button should route me back by layer to catalog but not to previous phone/tablet:
Image

4. Add the favicon and title to browser tab.

5. All links in Footer should lead to proper existing resourses:
Image

6. There should be no scroll on empty page, chcek same on shopping cart:
Image

7. Nice to have not less items in pagination than grid columns value, so remove 2 and set 4 as the smallest quantity:
Image

8. Also, there should be no scroll when its only 4 products in view and pagination should be sticky at the bottom of the page.

roman-mirzoian (changes requested):
Great job! Almost all of the previous issues have been fixed, except for point one from the previous review (problem with details page). Please check to see what the problem might be.

Image

Rostyslav452:
Okay, I get it. I use Firefox, and this problem doesn't appear. This layout bug seems browser-specific; I'm seeing it in Chrome, but everything works as expected in Firefox.
изображение

Anton-Kuchmasov (changes requested):
Cart page doesn't work as expected

Image

Would you mind to fix it?

here you can see an additional info that can be useful for debugging

Image

2pasha (changes requested):
good job! 👏

here are some improvements: 

1. it would be great to remove from cart item by clicking here 

Image

2. add some visual effect for active page 

Image

3. this button doesn't work 

Image

4. broken image

Image

5. broken 

Image

6. it will be great to open Product Details Page by clicking here

Image

roman-mirzoian (approved):
Well done!

PR #1040, opened 2026-07-14
solution

Oskar274:


brespect (changes requested):
Good progress, but you need to pass all tests and share your working DEMO LINK to the PR description.

Oskar274:
https://oskar274.github.io/react_phone-catalog/

etojeDenys (changes requested):
good job, to improve:
1. these should links to the relevant pages
Image

2. header should be slightly larger on the desktop version

Image

3. there should be more padding. make sure you set a maximum width for the page content

4. add a smooth scroll behaviour

5. after selecting a product, the page should scroll up

6. looks broken (also 'add to cart' button does not work)

Image

7. you should save these values in the search parameters so that they will not reset after refreshing the page

Image

Image

8. add an active style to the selected page number

9. don't forget this part

Image

10. re-check the checklist

roman-mirzoian (changes requested):
Good progress, but there are still a few points for improvement:
1. Make sure that all clickable elements have the `cursor: pointer` property. Currently, many elements do not have this property: buttons, cards, and pagination.
2. The initial banner should be smaller in size and include buttons for switching between items.
Image
3. The “Oder now” button on the banner must be a real button and link to the page for the specified product. 
Image
4. Favicon needs to be added.

Denys-Kravchuk9988 (changes requested):
Good job!

A few things to improve:
1. Waste symbol below the HTML page
Image

2. These tabs should have less width according to design
Image
Image

3. It's better to establish max-width for content according to design
Image
Image

4. Style of the button should be changed if it's clicked
Image
Image

5. By opening the new page is better to implement scrolling to top. 

6. Burger button looks broken
Image

7. Icons look broken on the opened menu
Image

8. It's better to implement slider on main page as on design
Image
Image

roman-mirzoian (approved):
Well done 👍

PR #1043, opened 2026-07-18
Phone Catalog

Inna-code10:


roman-mirzoian (changes requested):
When I click the link to the deployed project, an app that barely works opens. Please check the deployed version and fix the current issues:
Image

https://inna-code10.github.io/react_phone-catalog/

etojeDenys (changes requested):
good job, to improve:
Image

2. some products could not be found
Image

3. after selecting a product from this section, the page should scroll up
Image

4. after reloading a page that is not the home page, an error appears

Image

5. it would be good if the card could be removed from the cart by pressing the button again

Image

6. footer should not stretch
Image

Anton-Kuchmasov (approved):
Awesome!

PR #1047, opened 2026-07-21
task solution

droopy-bit:
[Demo Link](https://droopy-bit.github.io/react_phone-catalog/)


2pasha (changes requested):
good job! 👏

here are some improvements: 

1. product card should be the same height

Image

2. it would be great to remove item from cart by clicking here 

Image

3. no need to display all pages. Will be enough `1, 2, 3 .... 31` or smth like that

Image

4. it's better to use icons in `svg` format

Image

Denys-Kravchuk9988 (changes requested):
Good job! 

A few things to improve:
1. It's better to reset timer when slider's sticks are clicked. Now when I wait for 4-5 seconds and click next button and next slider is skipped and scrolled one more time.
Image

2. Brand new section shouldn't have discounts
Image

3. It's better to align pagination to center
Image

4. It's better to use `.svg` for icons and logos for better quality
Image

5. On tablet there should be 3 cards in row
Image
Image

2pasha (approved):
lgtm 🔥

PR #1048, opened 2026-07-21
Develop

chumachenkoUA:
[DEMO](https://chumachenkoua.github.io/react_phone-catalog/)


2pasha (changes requested):
good job! 👏

here are some improvements: 

1. items here should be without discounts 

Image

2. it would be great to remove item from cart by clicking here 

Image

2pasha (approved):
lgtm 🔥

PR #1050, opened 2026-07-24
solution

Alternate001:
[DEMO LINK](https://Alternate001.github.io/react_phone-catalog/)


brespect (changes requested):
1. Add the favicon:
Image

2. Can't find "Back" button near to bread crumbs:
Image

3. Add some padding around the text:
Image
Image

4. When I select something in "You may also like" section, the page should smoothly scrolls to the top:
Image

5. All links in Footer should be correct refs to existing resources:
Image

Image

7. Don't add the pagination button if there only one item on page:
Image

8. All products in this section should be without discounts:
Image

brespect (changes requested):
Good progress, but check next:
1. Don't add Mate as refs to Contact Us and Rights, better to use your own social medias and Apple Policy as example.

2. Back button should back me by layer as example to catalog, but not to previous phone/tablet.

3. There should be no scroll on empty page:
Image

4. Add hover to Product Card component.

roman-mirzoian (changes requested):
Great job! There are just a few corrections left:
- On the home page, when hovering over the first or last item, the product card extends beyond the border, so the outer corners of the card are hidden and disappear
Image
Image

- Add a `cursor: pointer` for the “Add to Cart” and “Favorites” buttons
- Add a link to your GitHub in the footer. Right now, there's a link to Mate's GitHub there.
Image

brespect (changes requested):
Good progress, check next:
1. There is some dismiss in logic, when I select phone with 256GB RAM and after that I select 512GB, the 512GB Phone Item will be shown as already added one, which is wrong because this is completely different one and I don't add it before:
Image

2. Also, as you can notice this item is already "Added to the Cart", but I can't see any counter badge in header next to shopping cart icon and inside this page:


3. There is no reason to show filters on empty favourite page:
Image

roman-mirzoian (approved):
Well done 👍

PR #1053, opened 2026-07-30
add task solution

ratushnjakivanko-spec:


brespect (changes requested):
Good progress, but you need to provide working DEMO LINK to the PR description, before requesting the review

ratushnjakivanko-spec:
https://ratushnjakivanko-spec.github.io/react_phone-catalog/

vadiimvooo (changes requested):
There are some problems with loading products and images. Also add favicon and title. Please feel free to ask for help in fe_chat.

roman-mirzoian (changes requested):
There's still some kind of loading issue with the demo:
Image

Also, add the vaficon and the page title instead of “Vite + React + TS”:
Image

brespect (changes requested):
Good progress, le'ts check next:
1. There is no "bread crumbs" icon:
Image

2. "Back" button should ref me back by layer to the catalog, not the previous item.

3. When I select another color/capacity or different item it should not reload the page.

4. When I select something in "You may also like" section it should smoothly scrolls the page to the top.
Image

5. All links in Footer must lead to proper existing resourses:
Image

6. Header should be fullwidth without left/right paddings.

7. Underline selection effect looks different from the Figma mockup:
Image

8. Mobile Header UI looks broken.

roman-mirzoian (approved):
Well done 👍

PR #1056, opened 2026-08-04
Develop

aamaltseva-sqwosh:


roman-mirzoian (changes requested):
Great start!
Here's a list of issues to improve:
- Add a favicon and a title for the browser tab:
Image

- Check that all images are linked correctly and exist, because some of them aren't loading right now:
Image

- In the “Hot prices” section, the image alignment isn't consistent, so hovering over larger images hides part of the image:
Image

- The active element “capacity” completely covers the value in the element, but it should be clear that it says:
Image

- Hovering over the product page may also hide part of the image:
Image

etojeDenys (changes requested):
almost done

1. in this section, products should not be discounted

Image

2. page number should be kept as it is for other parameters after a page reload (but it resets now)

Image

3. disable page scrolling when the menu is open

Image

roman-mirzoian (approved):
Well done 👍

PR #1057, opened 2026-08-04
PR №1

PavloMolytovnyk:
https://pavlomolytovnyk.github.io/react_phone-catalog/


brespect (changes requested):
Good progress, check next:
1. This Icons should be smaller:
Image

2. Right now, "Order Now" button do nothing and has different position from the mockup:
Image

3. When I select RAM size or color it should now reload the whole page:
Image

4. Add the favicon.

5. Every item that I select here route me to "Not found", also there is only 8 items, as when I use the scroll it shows me inly 3 more, so let's show all available items:
Image

6. Footer is different from the mockup:
Image

7. When I select something in "You may also like section" it should scroll me to the top of the page.

8. Don't forget to pass all tests and linter checks before requesting next review.

PavloMolytovnyk:
Hello, I fixed files, but dont understand 3 item: "When I select RAM size or color it should now reload the whole page:". Page is reloading after change color or ram. Maybe I dont good understand )))

Please, re-review the PR. Thanks

roman-mirzoian (changes requested):
It's a good start on the fixes, but there are still some differences with the design. There's no requirement to make the layout pixel-perfect, but the visual result should look close to the design:
- These buttons should be centered within their containers, and their borders should align with the header's border (see the design for reference):
Image

- The button and "Order now" container:
Current version:
Image

By design:
Image

- The buttons must be aligned either vertically or horizontally:
Image

- On the homepage, the prices on the product cards are huge, but on the page listing all products, they're lower. You need to standardize this to a single size, and I would recommend using the size from the design. I would also make the hover effect more subtle and faster:
Image
Image

- I'm sure the previous reviewer meant that the page “shouldn't” reload, but only the product itself should be refreshed:
Image

- On the product page, the active “Capacity” button changes its position relative to the other options:
Image

- All headings should start with a capital letter and should not wrap to the next line. Also, take a look at the rest of the text on this page’s design—since it’s not a heading, it’s not bold and is easier to read. I’d also recommend removing the bold formatting wherever it’s not used in the design:
Image

PavloMolytovnyk:
Hello, i fixed files, please re-review my PR

roman-mirzoian (changes requested):
I see that new code has been added, but the last deployment of your page was two days ago - so for now, I can't visually verify whether all the fixes have actually been applied. Please update the deployment 🙃 
Image

I also notice that you're using generic names for your commits - try to briefly describe exactly what your commit does. 
For example: `fix: center header buttons` or `feat: product card price font size`
where `fix` - a quick indicator that something that wasn't working properly has been fixed, and `feat` (feature) - a quick indicator that something has been updated or replaced with something new.
Image

PavloMolytovnyk:
i forgot to deploy)))

Now, the page must be changed

roman-mirzoian (changes requested):
Now the demo has been updated, and I can see the result :)

As I mentioned in my previous review, we're not trying to make it pixel-perfect, but the result should be similar to the design, since it was specifically designed to make the app look good
Here's what needs to be worked on:
- Take a look at the top of the homepage in the design: the welcome message with the app's name and the banner with the order button are wider than what you have now. You need to make it look like the design:
Image

- Check the text sizes in design. For example, the text size for prices on the cards is 22 pixels, but yours is set to 30 pixels, and the font looks too large.
- All hover enlargements are currently the same size and need to be reduced. This effect should subtly highlight the element and indicate to the user that they have hovered over it and can click on it.
- You aligned the icons in the header and gave them the correct border, but the hover effect adds extra borders, which causes the elements to "jump around":
https://gyazo.com/0cf5d2a5cc327f747fb23f0144cbb5f0

PavloMolytovnyk:
Fixed, as you notice ;)

PavloMolytovnyk:
I dont change hover on category images, because as written in the assignment "Scale all image links by 10% on hover."

etojeDenys (changes requested):
almost done, to improve:

Image

2. it would be better to make the card the same width

Image

3. center loader

Image

4. it should link to the product details

Image

PavloMolytovnyk:
fixed)

roman-mirzoian:
> I dont change hover on category images, because as written in the assignment "Scale all image links by 10% on hover."

You wrote the correct condition yourself - you need to scale the image itself, not the entire product card :)
However, the new hover animation is awesome 🔥

roman-mirzoian (approved):
Well done 👍

PR #1058, opened 2026-08-05
commit message

alexsand-r:
https://github.com/alexsand-r/react_phone-catalog


roman-mirzoian (changes requested):
Good start!
Here are a few issues to be fixed:
- Check to make sure the icons and images are linked correctly, because not everything is showing up in the demo right now:
Image

- The product page doesn't open the product; instead, it displays an error:
Image

- Add a favicon to the page
- Stick to a single color palette: right now, the main background is dark, the text is dark on a dark background, and the header and individual card backgrounds are white:
Image

roman-mirzoian (changes requested):
Great fixes, your app is working as expected!
But the pages for specific products still won't open. Please check the demo link to make sure everything is working correctly, and then submit it for review. Here is the correct link to the demo (correct the current link in the comment on the PR):
https://alexsand-r.github.io/react_phone-catalog/

Also, not all GitHub tests are passing right now; this needs to be fixed as well

alexsand-r:
https://alexsand-r.github.io/react_phone-catalog/

roman-mirzoian (approved):
Well done 👍

PR #1061, opened 2026-08-09
Finalize Nice Gadgets catalog

eg3844540-svg:
DEMO LINK:
https://eg3844540-svg.github.io/react_phone-catalog/


raulriato (changes requested):
You didn't provide a `DEMO LINK` and the one from gh-pages is not working.

Please make sure it works before re-requesting review, so the mentors are able to review your project.

src/modules/HomePage/components/Categories/Categories.module.scss
```diff
@@ -0,0 +1,65 @@
...
+.image {
+  display: block;
+
+  width: 100%;
+  aspect-ratio: 1;
+
+  object-fit: cover;
+
+  transition: transform 0.3s ease;
+}
```
raulriato, line 40: You should reposition the images here, and maybe, also add a background color to them, so they look neat.

src/components/ProductsSlider/ProductsSlider.module.scss
```diff
@@ -0,0 +1,55 @@
...
+.viewport {
+  overflow: hidden;
+}
+
+.products {
+  display: flex;
+  gap: 16px;
+
+  transition: transform 0.4s ease;
+}
```
raulriato, line 55: The facth you didn't add a `width` to this class is making the cards smaller than they should be. Also, after you fix this, you will probably need to add a `box-sizing` to the cards

raulriato (changes requested):
The main problems are pointed out directly in the files they belong to.

You should also implement the right `font-femily` to the project.

If you want to, it would be nice to add a transition effect to the slider in the `Home page`.

You are almost there, keep up.

DanielCamposDevX (changes requested):
-Images on buy page not rendering correctly + your back button is saying "back back"
- Order now button not working.

Remember to commit re-deploy and push your changes

src/modules/ProductDetailsPage/ProductDetailsPage.tsx
raulriato, line 1: The images are broken in the details page because you are fetching the `products.json` here. You should not do that, because `products.json` only have 1 image, not an array of images. Instead, you should fetch the specific product and category you are using in the page.

raulriato (changes requested):
You are almost there.

src/modules/ProductDetailsPage/ProductDetailsPage.tsx
```diff
@@ -125,7 +125,9 @@ export const ProductDetailsPage = () => {
     );
   }
 
-  const currentProduct = products.find(item => item.itemId === product.id);
+  const currentProduct = products.find(
+    item => item.category === category && item.itemId === product.id,
+  );
```
raulriato, line 108: The problem is not here. The problem is that you are fetching `products.json` to get the item you want to display. You need to fetch the `{category}.json` instead, so you get the details of the product. Also, you will use both `category` and `itemId` to `fetch` the product, so you can fetch only the product you need, instead of getting a full array and waste most of the items.

raulriato (changes requested):

DanielCamposDevX (approved):

PR #1063, opened 2026-08-11
add task solution

A-Melnykk:
[Product Catalog Demo](https://A-Melnykk.github.io/react_phone-catalog/)


etojeDenys (changes requested):
good job,  to improve:

Image

2. detail page does not work

Image

3. fix the logo image

4. make sure these buttons work

Image

5. page broken on the phone version

6. add a title and favicon to the page

Image

brespect (changes requested):
Sorry, but comments from previous @etojeDenys review were not fixed: logo, not found page, favicon & title etc. Please, request only when you will be ready

roman-mirzoian (changes requested):
At the moment, the changes aren't visible yet; please check to see if there might be any issues with the deployment. If you have any questions, message us in the chat and we'll work through them together:
Image

Anton-Kuchmasov (changes requested):
Great work you did!

To improve:
1. On `hover` phone image is looking cutted.
would you mind to fix this issue? Think about using `z-index` with required value here

Image

2. When I'm trying to open any Product page, it leads me to 404

Image

Probably, there's something wrong with URL?

Zoolgrand (changes requested):
Image
Looks like the links are broken, all catogiries empty with a bunch of 404 errors in the console

etojeDenys (changes requested):

Image

2. this one does not look like the actual phone color

Image

3. make sure that this section looks like in the design
Image

figma:

Image

4. don't forget to implement this logic

Image

etojeDenys (changes requested):
almost done
1. there are still discounted products in this section

Image

2. make sure the specific colors display correctly in those circles

Image

3. it should link to the product detail page
Image

4. and cursor pointer here. also, it would be better if the background color was the same as the background color of the page

Image

5. add some logic to this button, for example to clear the cart, when it is pressed

Image

PR #1068, opened 2026-08-14
added implementation

Skoulskiy:
[DEMO PAGE](https://skoulskiy.github.io/react_phone-catalog)


brespect (changes requested):
Good progress, but your DEMO LINK is not working:
Image

Also, don't forget to pass all CI checks before requesting the review

etojeDenys (changes requested):
demo is working now, but there is a problem with the products

Image

Image

roman-mirzoian (changes requested):
At the moment, the changes aren't visible yet; please check to see if there might be any issues with the deployment. If you have any questions, message us in the chat and we'll work through them together.
Image

etojeDenys (changes requested):
Image

Image

brespect (changes requested):
Good progress, check next:
1. There should be maximum of 4 items in this section:
Image

2. When I select something in "You may also like" section, it shoud smoothly scrolls page up.

3. "Back" button, should ref me back by layer to catalog, not to previous item.

4. All links in Footer should lead to proper existing resources:
Image

5. Add success reset flow instead of alert banner:
Image

6. When I select a pagination element, it's better not to annoy the user by scrolling up:
Image

etojeDenys (approved):
lgtm

PR #1071, opened 2026-08-31
solution

Nika-Andriy:
- [DEMO LINK](https://nika-andriy.github.io/react_phone-catalog/)


roman-mirzoian (changes requested):
Good start!
A few points for improvement:
- Add favicon and the project name to the page title in browser
- The element with the “Order now” button overlaps part of the content on the image; take a look at how it's done in the design and create a similar structure:
Image

- Usually (and in design as well), active and clickable elements are rendered in a rich color, while inactive ones are rendered in a more muted shade. You've got it the other way around right now, and it's a little confusing:
Image

- The active states and hover states of the buttons are currently overlapping, you need to either increase the spacing between them or slightly reduce their active states:
Image

- A few comments on this category: 1) When you hover over the container, a general shadow is added that sits right up against the elements - it’s unnecessary; 2) When the general hover effect appears and you hover specifically over the image, the image extends beyond the container’s borders, which looks a bit off:
Image

- On the checkout page, the product photo isn't loading, but the item itself should be clickable and link to the product page:
Image

- The icon in the tab isn't loading:
Image

- “Eabout” and “Tech Specs” should be on the same level (see the design):
Image

roman-mirzoian (changes requested):
Good progress!
A few things more:
- The element with the “Order now” button overlaps part of the content on the image; take a look at how it's done in the design and create a similar structure:
Image

- The list of items in the shopping cart should be clickable and link to the product page, right now, it's just an image and text:
Image

- If you select the minimum number of items to display in the product list, the tab pages stretch to fill the entire screen. Please make a more compact version, as shown in the example below:
Image

Image

Zoolgrand (approved):
Good job, you have fixed all the mentioned issues.
I'll approve this changes.
One minor fix si to add cursor pointer to this slider arrows

Image

PR #1074, opened 2026-09-02
phone-catalog

lishvets:
https://lishvets.github.io/react_phone-catalog/


brespect (changes requested):
Good progress, check next:
1. Something went wrong when I click on Order Now button in Slider:
Image

2. Add favicon and change title in browser tab.

3. Phones/Tablets/Accessories are not available:
Image

4. Footer should be sticky at the bottom of the page.

5. When I change theme to black one, it must change the Header too:
Image

roman-mirzoian (changes requested):
Good job!
Here are a few points to work on:
- Add a favicon
- In sections like these, all cards should be the same height, and it is preferable that the cards fit completely horizontally within the available space rather than being cut off in pieces:
Image

- When you hover over a product card, the image enlarges correctly but doesn't fit inside the container:
Image

- List pagination isn't working:
https://gyazo.com/ce8135c61388c838a41bcf48988474cd

- Small photos must have the “pointer” cursor property:
Image

- On a blank page, the content should be arranged so that the footer is immediately visible at the bottom of the screen and does not require scrolling to view. This applies to the Favorites page and the Cart page:
Image

PR #1077, opened 2026-09-07
Develop

MarianaKlochnyk:


roman-mirzoian (changes requested):
Good start!
The project structure is already in place. 
However, it’s currently unstable: images aren’t loading, none of the pages are opening, and the product cards don’t link to the actual products. 
Please check that everything works properly using the demo link, and only then submit the work for review:
Image
Image
Image
Image
Image

src/components/Header/Header.tsx
```diff
@@ -0,0 +1,316 @@
...
+
+  if (!contextCart) {
+    throw new Error('ShoppingCard must be used inside CartProvider');
+  }
+
+  const { cartItems } = contextCart;
+
+  let totalLength = 0;
+
+  cartItems.map(item => (totalLength += item.quantity));
```
DmytroStadnik: `.map` is an array methid that returns a value(modified array), so you have to assign the result somewhere
In this case it's usage is wrong because you are trying to iterate through the array and take field's values from it
The right way to do it would be to use an `array.reduce` method. Please read about it's usage

Also, this will be recalculated on every render even if the value doesn't change. Please read about memoization and `useMemo `and make `totalLength `a `useMemo`

src/components/ProductCardSale/ProductCardSale.tsx
```diff
@@ -0,0 +1,131 @@
+import { Link } from 'react-router-dom';
+import styles from './ProductCardSale.module.scss';
+import { useContext } from 'react';
+import { FavoritesContext } from '../../context/FavoritesContext';
+import { CartContext } from '../../context/CartContext';
+
+export type Product = {
```
DmytroStadnik: You already have Product type declared in ProductCard.tsx
You should not create two objects with the same name
And also in this case this Product looks like an extended variant of that Product, which means you can and should reuse this Product in ProductCard

src/context/CartContext.tsx
```diff
@@ -0,0 +1,65 @@
+import { createContext, useEffect, useState } from 'react';
+
+import { type Product } from '../components/ProductCardSale';
```
DmytroStadnik: I see that Product type is reused a lot. In this case it is wrong to declare it in a react component file. Please create file src/types.ts and move the Product and all other reusable types there

src/pages/AccessoriesPage/AccessoriesPage.tsx
```diff
@@ -0,0 +1,270 @@
...
+import { Loader } from '../../components/Loader';
+import { ProductList } from '../../components/ProductList';
+import { Link, useSearchParams } from 'react-router-dom';
+
+export const AccessoriesPage = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const [accessories, setAccessories] = useState<Product[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [hasError, setHasError] = useState(false);
```
DmytroStadnik: Add generics to all useStates, even where type is obvious

src/pages/AccessoriesPage/AccessoriesPage.tsx
```diff
@@ -0,0 +1,270 @@
...
+    sortedAccessories.sort((a, b) => getYear(b) - getYear(a));
+  }
+
+  if (typePerPage === 'title') {
+    sortedAccessories.sort((a, b) => a.name.localeCompare(b.name));
+  }
+
+  if (typePerPage === 'price') {
+    sortedAccessories.sort((a, b) => a.priceDiscount - b.priceDiscount);
+  }
```
DmytroStadnik: Create a useMemo for sorted array, which will depend on typePerPage

src/pages/PhonesPage/PhonesPage.tsx
```diff
@@ -0,0 +1,267 @@
...
+  const [currentPage, setCurrentPage] = useState(
+    Number(searchParams.get('page')) || 1,
+  );
+  const [typePerPage, setTypePerPage] = useState(
+    searchParams.get('sort') || 'age',
+  );
+
+  const sortedPhones = [...phones];
+
+  const getYear = (product: Product) => {
```
DmytroStadnik: I think you already have this function declared in another component. If so, move this function to a function library like helpers.ts and import it where needed

DmytroStadnik (changes requested):
Nice progress. I see images and everything seems to work, but for me the fonts don't load(404). Please check this

Also, I've left a few comments in the code about potential issues and wrong usage of array methods and code reuse. Please check them and update the code

src/components/ProductCardSale/ProductCardSale.tsx
MarianaKlochnyk: I understand. I have two Product types because I use them for different things.
The Product type in ProductCard is used for the products from products.json on the Home page. The Product type in ProductCardSale is used for the full product data from phones, tablets and accessories, and I also use it in CartContext and FavoritesContext.
They have different fields and different id types, so I wasn't sure how to reuse one Product in the other.
Could you please clarify how you would recommend doing this?

roman-mirzoian (changes requested):
Good progress!
Here's what needs to be refined:
- Add a favicon and the website name:
Image

- The “Order now” button should be clickable and link to the product being offered:
Image

- A blank image of an empty iPhone looks as if there should be some content there, but it didn't load:
Image

- When hovering over product cards or images, there should be a zoom effect, as described in the assignment.
- The layout should be the same for all cards:
Image

- On the product page: the white icon has no border and is therefore invisible; when the color or other settings are changed, nothing changes (even though it should); the price has no space between the digits:
Image

- The layout is also gone:
Image

- On an empty shopping cart page, the footer should appear at the bottom:
Image

- If a product is selected - for some reason, the footer isn't at the bottom either, and the product card itself should be clickable and link to the product page:
Image

- There's no space here either, and the first line of parameters is out of place:
Image

- On a blank page, the Favorites footer should appear at the bottom of the screen:
Image

MarianaKlochnyk:
I'm not sure where the "Order now" button should link to. Could you please clarify?

2pasha (commented):
good job! 👏

here are some improvements: 

1. change favicon

Image

2. make sure that you are using right fonts over all your pages 

Image

3. items in this section should be without any discount 

Image

4. make all your animation/transition smooth (item hovering, banner scrolling and so on)

Image

5. add hover effects for these items 

Image

6. you do not need to display all pages. it's better to do something like this `1, 2, 3 ... 31`

Image

7. make sure that here is right margin

Image

8. add some real links for these items 

Image

2pasha:
> I'm not sure where the "Order now" button should link to. Could you please clarify?

you may make this as a link to your phone catalog for example or to some specific product details page

2pasha (changes requested):
please take a look on comments before

etojeDenys (changes requested):
good job, to improve:
1. slightly decrease the scale of the hovered image to avoid merging and overlapping

Image

2. price font looks different

Image

3. these buttons should be slightly higher

Image

4. make sure you show the correct color on the buttons (it does not look black)

Image

5. after selecting a product from this section, the page should scroll up
Image

6. hover effect should also be smooth here

Image

roman-mirzoian (changes requested):
Almost done!
- The hover effect causes the cards to extend beyond the container, causing the card borders to disappear:
Image

- In this section, only the image should expand, not the entire container:
Image

- The font used for prices varies from place to place; standardize everything to the font specified in the design:
Image
Image

etojeDenys (changes requested):
almost done
1. you forgot to change the font style here as well

Image

Image

2. make sure that this link opens in a new tab

Image

roman-mirzoian (approved):
Well done 🔥

PR #1082, opened 2026-09-10
Add project pages, components and shared modules

DmytroYakovchuk:


2pasha (changes requested):
project looks broken 

Image

roman-mirzoian (changes requested):
Good start!
Points for improvement:
- You need to add a favicon and the correct site name to the browser
- When you hover over a card, the image may be cropped, and the last card in the row does not fit entirely within the container and is also cropped:
Image

- These buttons should either be centered vertically or made the same height:
Image

- When the “Add to Cart” button is clicked, the button resizes and the card height shifts slightly. It would also be nice to be able to cancel the action of adding an item to the cart by clicking this button again.
- The GitHub link at the bottom of the page isn't working.
- The available colors aren't displaying properly:
Image

- Products on the page may appear with different layouts depending on their color:
Image
Image

- Pagination should not take up an entire page; it should look compact:
Image

- It's better to move the search field to a single line along with the sorting, and the arrow icons should be positioned with appropriate padding:
Image

- For blank pages, the footer should remain at the bottom and not move. This also applies to the Favorites and Cart pages:
Image

- It would be better to display the total amount of items in the cart as part of the design, because right now, if many items are selected, it's unclear that this element even exists:
Image

- Check that everything works as expected in the mobile version:
Image

Anton-Kuchmasov (changes requested):
Great work you did!

Just a few improvements need to be made:
1. Fix color bullets here:

Image

2. When Users selects one of products from this block, please make page scrolling smoother - now it works instantly:

Image

3. Products in the Cart should be clickable - click on each of this articles should redirect us to chosen Product' page:

Image

roman-mirzoian (changes requested):
Some of the issues I wrote about earlier still haven't been fixed:
- A change in color or any other device setting should appear seamless, but right now it's very noticeable because the container size changes:
Image

- Image
- Image
- Image

brespect (changes requested):
Good progress, check next:
1. Make header full width for wide desktop screens:
Image

2. Let's add some images to gallery swiper and implement "Order Now" button:
Image

3. Products in "Brand New Model" section should be without discounts:
Image

4. Seems like pagination is not working and it should be sticky at the bottom of the page: 
Image

5. "Back" button should ref me back by layer, to catalog, but not to previous item:
Image

6. Add "Bread Crumbs" icon here:
Image

7. There are a lot of yellow iphones in catalog, but when I search with prompt like "Apple Yellow" it shows me nothing:
Image

8. Check this naming, as it looks somewhat strange and disappointing to the user:
Image

PR #1085, opened 2026-09-16
react_phone-catalog

gadyka08-lab:
https://gadyka08-lab.github.io/react_phone-catalog


etojeDenys (changes requested):
demo link does not work. please fix it and re-request the review

2pasha (changes requested):
good job! 👏

here are some improvements: 

1. it's better to use svg for icons

Image

2. add counter for your favorites and cart

Image

3. fix these numbers 

Image

4. change title and favicon
5. fix product details page, because right now it looks like this 

Image

6. no need to display all pages, you may do this like this `1, 2, 3 ... 31`

Image

7. fix cart

Image

PR #1086, opened 2026-09-16
version 1

dianakomar:
https://dianakomar.github.io/react_phone-catalog/


brespect (changes requested):
Good progress, but your DEMO LINK is not working

brespect (changes requested):
Nothing changes, please request only when you will be ready

etojeDenys (changes requested):
good job, to improve:

Image

2. too much space here

Image

3. these links lead to an error page. it would be better to fix this

Image

4. the footer should not stretch

Image

5. add a favicon and title to the page
Image

PR #1088, opened 2026-09-17
Develop

nester02:
https://nester02.github.io/react_phone-catalog/#/


roman-mirzoian (changes requested):
Good start!
A few points for improvement:
- The initial carousel element looks too big, make it the same size as in the design:
Image
 
- After clicking the “Add to Cart” button, it would be nice to be able to remove the item from the cart by clicking that same button
- If the cart is empty, the footer should be visible right away, not after scrolling. This logic is already implemented on your Favorites page:
Image

roman-mirzoian (changes requested):
Good fixes!
But for some reason, the initial banner on the demo version isn't loading the images. Please check how everything works using the demo link and resubmit your work for review:
Image

roman-mirzoian (approved):
Well done 👍

PR #1093, opened 2026-09-22
Develop

kseniiavasylenko:
https://kseniiavasylenko.github.io/react_phone-catalog/


etojeDenys (changes requested):
good job, to improve:

Image

2. swap the places of these sections

Image

3. there is a double space here
Image

4. looks broken

Image

5. you should keep these values in search parameters

Image

Image

6. product page does not work

Image

7. fix icons

Image

Image

Zoolgrand (changes requested):
There is still some issues to fix

Cart page product thumbnail and price is broken

Image

Pagination shouldn't display all available options, also we need to save pagi in url search param

Image

brespect (changes requested):
Good progress, check next:
1. Pagination should be sticky at the bottom of the page:
Image

2. Add gap between Header and this Title: 
Image

3. When I add this item to the cart or favorites, it should be marked as added, and the heart's background should be filled in:
Image

4. When I select something in "You may also like" section it should smoothly scrolls the page to the top.

5. "Back" button in Bread Crumbs should ref me back by layer, to catalog, not to previous item.

roman-mirzoian (approved):
Well done 👍

